Issue #81, opened 2025-01-13
sortMeRNA fails


ulbivin:
### Description of the bug

I am trying to run nf-core/riboseq on a cluster computer, but it fails on the sortMeRNA step for all samples. 

### Command used and terminal output

```console
I run this nextflow command:
$ nextflow run nf-core/riboseq -profile computerome --project reg_00024 --input nf-core_riboseq_sample_sheet.csv --outdir /home/projects/reg_00024/data/RiboSeq/nf-core_riboseq_out/ --fasta /home/projects/reg_00024/data/RiboSeq/ref/Homo_sapiens.GRCh38.dna_sm.primary_assembly.fa.gz --gtf  /home/projects/reg_00024/data/RiboSeq/ref/Homo_sapiens.GRCh38.113.gtf.gz  --save_reference -resume

This is one of the error messages from the .nextflow.log file:
Jan-09 16:39:14.631 [Task monitor] DEBUG n.processor.TaskPollingMonitor - Task completed > TaskHandler[jobId: 51991560; id: 61; name: NFCORE_RIBOSEQ:RIBOSEQ:PREPROCESS_RNASEQ:SORTMERNA (0279_05_RiboSeq); st
atus: COMPLETED; exit: 137; error: -; workDir: /home/projects/reg_00024/data/RiboSeq/work/e1/24e9e6f1ba0dd9c49c905846f6bf53 started: 1736431257223; exited: 2025-01-09T15:39:04.322921661Z; ]
Jan-09 16:39:14.636 [TaskFinalizer-6] DEBUG nextflow.processor.TaskProcessor - Handling unexpected condition for
  task: name=NFCORE_RIBOSEQ:RIBOSEQ:PREPROCESS_RNASEQ:SORTMERNA (0279_05_RiboSeq); work-dir=/home/projects/reg_00024/data/RiboSeq/work/e1/24e9e6f1ba0dd9c49c905846f6bf53
  error [nextflow.exception.ProcessFailedException]: Process `NFCORE_RIBOSEQ:RIBOSEQ:PREPROCESS_RNASEQ:SORTMERNA (0279_05_RiboSeq)` terminated with an error exit status (137)
[...]

PR #105, opened 2025-04-30
SortMeRNA low mem params

JackCurragh:
…. Any rRNA hit is enough to characterise read as rRNA not RPF. This should lower memory usage and address some user reported errors on slack and [#81]. 

## PR checklist

- [x] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/riboseq/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, also make a PR on the nf-core/riboseq _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository.
- [x] Make sure your code lints (`nf-core pipelines lint`).
- [x] Ensure the test suite passes (`nextflow run . -profile test,docker --outdir <OUTDIR>`).
- [ ] Check for unexpected warnings in debug mode (`nextflow run . -profile debug,test,docker --outdir <OUTDIR>`).
- [ ] Usage Documentation in `docs/usage.md` is updated.
- [ ] Output Documentation in `docs/output.md` is updated.
- [x] `CHANGELOG.md` is updated.
- [ ] `README.md` is updated (including new tool citations and authors/contributors).


pinin4fjords (approved):
Sounds reasonable, approving. Perhaps we can see how well this works in this pipeline for a while, and then deploy in rnaseq if we don't have issues.

iraiosub:
Thanks for looking into this — `--no-best` does avoid the exhaustive search and ranking even with `--num_alignments 1`, which explains the memory savings. That said, this issue seems extremely rare, despite the widespread use of SortMeRNA in nf-core/rnaseq, and since `--no-best` changes alignment behavior, it might be better as a documented option for filtering or large datasets rather than a default change for everyone? It could also affect rRNA subtype reporting in MultiQC, so unless the performance gains turn out to be consistently big, it might be worth holding off on changing the default? It feels like a case where a config override would be the better path unless we start seeing this come up more frequently.

JackCurragh:
Hi @iraiosub, I take your point that rRNA subtype might change but my intuition (which is often wrong) would be that the high identity required for a alignment would mean we would be getting the right subtype regardless. I also checked with this branch versus `dev` using `-profile test` and the multiqc_sortmerna.txt files are identical. I dont have an alternative large dataset at hand currently to test this further.  

I also suspect the reason this has less frequently come up for nf-core/rnaseq is that the rRNA content is significantly different between these two data types. Also, perhaps this goes agsinst my point above but, since the reads are longer in RNA-Seq multimapping of rRNA reads is less common? 

Regardless, not going to merge until we are at an agreement. I can of course make this an option if its preferred but don't have a tonne of time at the moment.

pinin4fjords:
FWIW I do get queries along these lines fairly frequently for SortMeRNA in rnaseq

iraiosub:
Thanks for the thoughtful reply @JackCurragh! Just to clarify — this is already configurable via `-params-file <file>`, so users can enable `--no-best` themselves without any extra work on your/our end. I wasn’t suggesting making it optional within the pipeline itself. I’m not against merging if that’s preferred or it's a common issue, but if so, it would be helpful to document in the pipeline docs that `--no-best` is set by default because the main goal is rRNA filtering, and it improves performance — with the tradeoff that the reported alignment may not always be the best match.

PR #135, opened 2025-12-16
feat: Add optional read length equalisation for RNA-seq/Ribo-seq comparison

pinin4fjords:
## Summary

- Add optional feature to trim RNA-seq reads to match Ribo-seq read lengths before quantification
- Addresses potential biases when comparing the two modalities due to different read lengths
- Enabled via `--equalise_read_lengths` parameter

### New parameters
- `--equalise_read_lengths`: Enable the feature (default: false)
- `--equalise_read_lengths_target`: Global target length fallback

### New samplesheet columns
- `trim_length`: Per-sample target length override (accepts string or integer)
- `pair`: Sample pairing identifier for matching RNA-seq to Ribo-seq samples

### Implementation
- New local subworkflow `FASTQ_EQUALISE_READ_LENGTHS`
- Priority for trim length: per-sample `trim_length` > global target > derived from paired Ribo-seq
- Uses seqkit stats to measure Ribo-seq read lengths (only runs when derivation is needed)
- Uses TrimGalore `--hardtrim5` to trim RNA-seq reads from 5' end
- For paired-end RNA-seq, only R1 is retained (preserving 5' position)
- Integrates seqkit stats output with MultiQC (separate sections for rRNA removal and read length measurement)

### Documentation
- Updated `docs/usage.md` with comprehensive feature documentation
- Added CHANGELOG entry

## Test plan

- [x] Subworkflow nf-tests (paired samples, global target, per-sample trim_length, passthrough)
- [x] Pipeline-level nf-test for full integration
- [ ] Manual testing with real RNA-seq/Ribo-seq data


docs/usage.md
```diff
@@ -156,6 +157,59 @@ nextflow run nf-core/riboseq --ribo_removal_tool ribodetector ...
...
+   - Per-sample `trim_length` column in the samplesheet
+   - Global `--equalise_read_lengths_target` parameter
+   - Average length derived from the paired Ribo-seq sample (via `seqkit stats`)
+2. **Ribo-seq read length measurement**: `seqkit stats` is only run on Ribo-seq samples that have paired RNA-seq samples needing length derivation (i.e., when neither a per-sample `trim_length` nor global target is specified)
+3. **Trimming**: RNA-seq reads are hard-trimmed from the 5' end using TrimGalore's `--hardtrim5` option
+4. **Paired-end handling**: For paired-end RNA-seq, only R1 is retained (preserving 5' position information)
+
+If none of these sources provide a trim length for an RNA-seq sample, the pipeline will exit with an error.
+
+### When to use
```
JackCurragh, line 176: I think this does a great job explaining the relevance of the approach and also why you may not wish to use it. 

Lets just hope people read it..

assets/schema_input.json
```diff
@@ -38,6 +38,18 @@
                 "errorMessage": "Sample type must be provided and be one of 'riboseq', 'rnaseq' or 'tiseq'",
                 "enum": ["riboseq", "rnaseq", "tiseq"],
                 "meta": ["sample_type"]
+            },
+            "trim_length": {
+                "anyOf": [
+                    { "type": "integer", "minimum": 15 },
+                    { "type": "string", "pattern": "^[0-9]+$" }
+                ],
+                "errorMessage": "Trim length must be a positive integer of at least 15",
```
JackCurragh, line 47: I'm not sure I understand why we need a min value here. I guess nobody should be bothered to use this if the lenght difference is less than 15.

workflows/riboseq/main.nf
```diff
@@ -168,13 +169,43 @@ workflow RIBOSEQ {
...
+    // SUBWORKFLOW: Equalise RNA-seq read lengths to match Ribo-seq read lengths
+    //
+
+    // Normalize samplesheet-derived meta fields (convert empty lists to null)
+    // Only modify fields that exist in the meta
+    ch_reads_preprocessed = FASTQ_QC_TRIM_FILTER_SETSTRANDEDNESS.out.reads
+        .map { meta, reads ->
+            def updates = [:]
+            if (meta.containsKey('trim_length')) {
+                updates.trim_length = meta.trim_length instanceof List ? (meta.trim_length ? meta.trim_length[0] : null) : meta.trim_length
```
JackCurragh, line 182: Would you mind explaining how this could end up being a List? I'm struggling to follow

subworkflows/local/fastq_equalise_read_lengths/main.nf
```diff
@@ -0,0 +1,156 @@
...
+                return [ meta, reads ]
+            other: true
+                return [ meta, reads ]
+        }
+        .set { ch_reads_by_type }
+
+    // Determine trim length for each rnaseq sample based on priority:
+    // 1. Per-sample samplesheet trim_length on the RNA-seq sample itself
+    // 2. Global equalise_read_lengths_target parameter
+    // 3. Derived from paired Ribo-seq via seqkit stats (only if needed)
```
JackCurragh, line 42: Why is this approach such low priority? It seems like matching the read length distribution this was is the only truly correct approach? Singular global RNA-Seq legnths would likely have its own biases. 

I'm not sure how important it is to get into the weeds on this though..

JackCurragh (approved):
I believe I could follow the workflow logic and it all seems sound. 

My one concern is, if we are to give this as an option then perhaps the only approach we offer should be to apply this to paired samples and based on the pairing match the read length distribution of the RNA-Seq to the Ribo-Seq. Otherwise we are just moving the bias rather than removing it. 

But at the same time, the docs and code are clear about what is done and if this is the design that @FelixKrueger and his team are happy with then I am happy to give a thumbs up. I do think it will mostly null any mapping effects on count differences

subworkflows/local/fastq_equalise_read_lengths/main.nf
pinin4fjords: It's not priority in terms of merit, just in terms of implementation. If someone has explicitly set a length they want to use (either sample-wise or global) it should prevent the auto-computation. I still expect that to be the primary way.

JackCurragh: Makes sense. Thanks

workflows/riboseq/main.nf
pinin4fjords: When samplesheet columns are empty, nf-schema returns empty lists rather than null:

Samplesheet:

```
sample,fastq_1,strandedness,type,trim_length,pair
test1,test1.fastq.gz,forward,riboseq,,
test2,test2.fastq.gz,forward,rnaseq,28,1
```

Result in meta:

test1: trim_length: [] (ArrayList), pair: [] (ArrayList)
test2: trim_length: 28 (Integer),   pair: 1 (Integer)

So the helper function normalises both cases - extracting the value from a list if present, or returning null for empty lists.

assets/schema_input.json
pinin4fjords: It's just so the users get a warning if they do something silly

workflows/riboseq/main.nf
JackCurragh: Good to know! Thank you once again for the lesson

PR #137, opened 2025-12-17
feat: Add Salmon pseudo-alignment pathway for TE analysis

pinin4fjords:
## Summary

Add an alternative quantification method for translational efficiency (TE) analysis using direct Salmon pseudo-alignment instead of the default STAR→Salmon alignment-based mode.

### Motivation

When analyzing TE (Ribo-seq / RNA-seq ratio), different read lengths between modalities can introduce mappability bias - short ~30nt Ribo-seq reads and long ~150nt RNA-seq reads map to different "effective transcriptomes". The existing `--equalise_read_lengths` feature addresses this by trimming RNA-seq reads, but this discards sequence information.

This PR provides an alternative approach: running Salmon pseudo-alignment for both modalities using the same k-mer index, putting them on equal footing without trimming.

### New Parameters

| Parameter | Default | Description |
|-----------|---------|-------------|
| `--te_quantification_method` | `alignment` | Choose `alignment` (STAR→Salmon) or `pseudo` (Salmon pseudo-alignment) |
| `--pseudo_aligner_kmer_size` | `23` | K-mer size for Salmon index (optimized for short Ribo-seq reads) |

### How it works

When `--te_quantification_method pseudo` is set:

1. A Salmon index is built with k-mer size 23 (default, optimized for short Ribo-seq reads)
2. Ribo-seq and RNA-seq reads go through Salmon pseudo-alignment directly
3. The resulting counts are routed to deltaTE/anota2seq
4. STAR alignment still runs for position-dependent analyses (P-sites, periodicity, ORF detection)

### Usage

```bash
nextflow run nf-core/riboseq \
    --te_quantification_method pseudo \
    --contrasts contrasts.csv \
    ...
```

## Test plan

- [x] Run test with `--te_quantification_method pseudo`
- [x] Verify Salmon index is built with k=23
- [x] Verify pseudo counts are used for TE analysis
- [x] Verify STAR alignment still runs


conf/modules.config
```diff
@@ -68,6 +68,18 @@ process {
         ]
     }
 
+    withName: 'SALMON_INDEX_TE' {
+        ext.args   = { [
+                params.gencode ? '--gencode' : '',
+                params.pseudo_aligner_kmer_size ? "-k ${params.pseudo_aligner_kmer_size}" : ''
```
suhrig, line 74: You PR description says you introduce a new parameter `--te_pseudo_kmer_size` for this task. Did you abandon this idea and prefer to use the same parameter for both regular Salmon quant and TE Salmon quant?

(If you meant to use `--te_pseudo_kmer_size`, then `--te_pseudo_aligner_kmer_size` may be the better name for the sake of consistency.)

suhrig (commented):
Apart from the discrepancy regarding `--te_pseudo_kmer_size` in the code and your PR description, the PR looks good to me.

conf/modules.config
pinin4fjords: Sorry, I forgot to tidy up a bit.

Basically I realised that the current workflow doesn't use the kmer size for anything much- we pass BAM files to salmon for its 'aligment mode' which doesn't consume a Salmon index (which is what the kmer size is used for).

So a new param didn't make sense, I just co-opted the old one.

suhrig: That's what I suspected. Just wanted to double-check.

suhrig (approved):

PR #134, opened 2025-12-16
Plastid

suhrig:
This PR incorporates the new plastid module into the pipeline. This is the first step to making quantification of Ribo-seq data more accurate. It enables us to quantify Ribo-seq reads which are in-frame instead of counting the raw reads.

The second step will be to incorporate the counts into the downstream analyses. I will wait with this until the current improvements around the quantification scripts have been completed (@pinin4fjords is currently working on this AFAIK).

## PR checklist

- [X] This comment contains a description of changes (with reason).
- [X] If you've fixed a bug or added code that should be tested, add tests!
- [X] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/riboseq/tree/master/.github/CONTRIBUTING.md)
- [X] Make sure your code lints (`nf-core pipelines lint`).
- [X] Ensure the test suite passes (`nextflow run . -profile test,docker --outdir <OUTDIR>`).
- [x] Check for unexpected warnings in debug mode (`nextflow run . -profile debug,test,docker --outdir <OUTDIR>`).
- [X] Usage Documentation in `docs/usage.md` is updated.
- [X] Output Documentation in `docs/output.md` is updated.
- [X] `CHANGELOG.md` is updated.
- [X] `README.md` is updated (including new tool citations and authors/contributors).

PR #140, opened 2026-01-13
Plastid

suhrig:
This PR is identical to #134. I accidentally deleted the latter assuming it had been merged already. The content of this PR is identical (as can be confirmed by the commit hash `49a1849`). The only thing I changed on top is a comment in the CHANGELOG, which needed updating, because the old PR number was referenced there.


pinin4fjords (approved):

suhrig:
Many thanks for your support, @pinin4fjords!

PR #141, opened 2026-01-18
in-frame p-site quantification method

suhrig:
This PR adds another (experimental) quantification method for Ribo-seq data. Instead of counting all Ribo-seq reads which map to a transcript, this new method only counts reads which map to coding regions (`CDS`) and only if the P-sites computed by Plastid align with the reading frame of the coding region. This means the quantification method is fairly conservative, ignoring frame-shifted reads or reads mapping to unannotated coding regions. RNA-seq quantification is unchanged.

The user can enable this quantification method by setting `--te_quantification_method` to the value `plastid_psite`. It is off by default. The method is implemented as a local module (essentially `bedtools intersect`).

I meant to add a test for this new quantification method, but the inputs to the module are not deterministic to begin with which means the outputs aren't either, making a test rather pointless.

## PR checklist

- [x] This comment contains a description of changes (with reason).
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [x] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/riboseq/tree/master/.github/CONTRIBUTING.md)
- [x] Make sure your code lints (`nf-core pipelines lint`).
- [x] Ensure the test suite passes (`nextflow run . -profile test,docker --outdir <OUTDIR>`).
- [x] Check for unexpected warnings in debug mode (`nextflow run . -profile debug,test,docker --outdir <OUTDIR>`).
- [x] Usage Documentation in `docs/usage.md` is updated.
- [x] Output Documentation in `docs/output.md` is updated.
- [x] `CHANGELOG.md` is updated.
- [x] `README.md` is updated (including new tool citations and authors/contributors).


suhrig:
@nf-core-bot fix linting

docs/usage.md
```diff
@@ -395,6 +395,23 @@ Consider this option when:
...
 
+#### In-frame P-sites
+
+```bash
+--te_quantification_method plastid_psite
+```
+
+In this mode, the RNA-seq reads are quantified using the default alignment-based quantification and the Ribo-seq reads are counted only when they map to coding regions and only when their P-sites (as determined by Plastid) coincide with the reading frame of the coding regions. This is an **experimental mode** which applies conservative filters to Ribo-seq quantification. It has applications in the study of microproteins arising from non-canonical ORFs, where an abundance of in-frame P-site counts increases the confidence that a novel ORF call is reliable. It is less common in other applications of Ribo-seq. Consider this option when:
+
+- Your use case is in the field of detecting non-canonical ORFs coding for microproteins
```
JackCurragh:
```suggestion
- Your use case is in the field of detecting differential translation of non-canonical ORFs 
```

There is nothing that makes it microprotein specific..

modules/local/quantify_inframe_psite/main.nf
```diff
@@ -0,0 +1,95 @@
...
+                    header[i] = \$i
+            } else {
+                # replace values whenever we find a column of a Ribo-seq sample
+                for (i = 3; i <= NF; i++)
+                    if (header[i] in samples)
+                        \$i = count[header[i], \$1] + 0
+            }
+            print
+        }
+    ' /dev/stdin ${counts} > "${feature}_counts_length_scaled.tsv"
```
JackCurragh: Is this really length scaled? I may well be missing something but it looks to me to just retunr counts?

JackCurragh (changes requested):
Minor wording update and just requesting clarification on the counting method. Can you point me to the line that length-scales the counts?

modules/local/quantify_inframe_psite/main.nf
suhrig: Well-spotted. It's not. I used to have this length-scaled in an earlier version, but now it's just raw counts. I will fix this.

iraiosub (commented):
- The docs should make it much clearer that only annotated CDS in-frame P-sites are counted. As written, there is a long and repetitive discussion of non-canonical and microprotein translation use cases, which (for a feature that only counts over canonical CDS) I believe should be shortened substantially
- Is there a reason only Plastid P-sites are being used for this quantification? Since the pipeline supports multiple P-site callers, should the others be an option too?

suhrig:
> The docs should make it much clearer that only annotated CDS in-frame P-sites are counted. As written, there is a long and repetitive discussion of non-canonical and microprotein translation use cases, which (for a feature that only counts over canonical CDS) I believe should be shortened substantially

I have removed the references to microproteins and made the description more explicit.

> Is there a reason only Plastid P-sites are being used for this quantification? Since the pipeline supports multiple P-site callers, should the others be an option too?

No particular reason other than I was most familiar with this tool. I can add Ribowaltz as an alternative option. Any others?

iraiosub:
- thanks! riboWaltz would be useful. it natively outputs the in-frame CDS counts in this file `*.cds_coverage_psite.tsv`,  so the implementation only needs a new param, and wiring these counts into the TE (i.e. there is no need to compute in-frame counts all over)
- just saw the comment re testing, if you're still keen to add it, it is still possible with unstable outputs. You'd only need to either exclude this process's outputs from the snapshot, or test existence without snapshotting. In this case, I think testing is useful because these changes include a local module

suhrig:
> riboWaltz would be useful. it natively outputs the in-frame CDS counts

I tried. However, the whole pipeline is currently build around genes, wheres riboWaltz is [hard-wired to use transcripts](https://github.com/nf-core/riboseq/blob/52244e56447fdfce9efe7c005a385ac7272a5ec3/modules/nf-core/ribowaltz/templates/ribowaltz.r#L110). Either the pipeline needs to be enhanced to work with transcripts, too, or riboWaltz needs to be enabled to deal with counts aggregated by genes. Probably both. In any case, those are changes that need to happen independently from this PR.

What I have done for now is to split the process into counting p-sites and editing the counts matrix. This way, when we want to apply other tools analogously to plastid, they can be wired into the workflow easily.

> just saw the comment re testing, if you're still keen to add it, it is still possible with unstable outputs.

I have added a test for the module.

suhrig:
@nf-core-bot fix linting

suhrig:
@iraiosub As explained above, fitting riboWaltz into the workflow is currently not possible. Do you think we can limit the support to plastid for now and add riboWaltz in a separate PR once the obstacles have been moved out of the way?

modules/local/quantify_inframe_psite_plastid/main.nf
```diff
@@ -0,0 +1,66 @@
...
+    input:
+    tuple val(id), path(forward_bedgraph), path(reverse_bedgraph)
+    tuple val(meta), path(gtf)
+    val feature
+
+    output:
+    tuple val(meta), path("*_inframe_psite_counts.tsv"), emit: counts
+    path "versions.yml", emit: versions
+
+    script:
```
pinin4fjords, line 23: This module should have a stub

modules/local/quantify_inframe_psite_plastid/main.nf
```diff
@@ -0,0 +1,66 @@
...
+        bedtools intersect -wa -wb -loj -a psites.bed -b /dev/stdin -s |
+        awk -F'\\t' -v OFS='\\t' -v SAMPLE="\$SAMPLE" '{ print SAMPLE, \$4, int(\$10) }' |
+        bedtools groupby -g 1,2 -c 3 -o sum
+
+    done > "${feature}_inframe_psite_counts.tsv"
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        bedtools: \$(bedtools --version | cut -f2 -dv)
+        awk: \$(awk 2>&1 | head -n1 | cut -f2 -d' ' | cut -f2 -dv)
```
pinin4fjords: I think this might behave differently depending on environment. I would suggest you actually declare gnu awk as a dependency like https://github.com/nf-core/modules/blob/master/modules/nf-core/star/align/environment.yml, so you can be more confident.

modules/local/quantify_inframe_psite_plastid/main.nf
```diff
@@ -0,0 +1,66 @@
...
+            # print coordinates of codon starts
+            frame = strand == "+" ? phase : (3 - phase + end - start) % 3
+            for (pos = start + frame; pos <= end; pos += 3)
+                   print chrom, pos-1, pos, feature, 0, strand
+        }
+    ' "$gtf" |
+    sort -u > psites.bed
+
+    # collate sample IDs and bedgraph files
+    echo -e '${ [id, forward_bedgraph, reverse_bedgraph].transpose().collect { it.join('\\t') }.join('\\n') }' |
```
pinin4fjords: This logic is effectively re-implementing something Nextflow could do (which would scale better). You're not doing any cross-sample comparison as far as I can see, so there's no need to do it this way.

modules/local/quantify_inframe_psite_plastid/main.nf
```diff
@@ -0,0 +1,66 @@
...
+    tag "$gtf"
+    label 'process_single'
+
+    conda "bioconda::bedtools=2.31.1"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/bedtools==2.31.1--h13024bc_3' :
+        'biocontainers/bedtools:2.31.1--h13024bc_3' }"
+
+    input:
+    tuple val(id), path(forward_bedgraph), path(reverse_bedgraph)
```
pinin4fjords:
```suggestion
    tuple val(meta), path(forward_bedgraph), path(reverse_bedgraph)
```

Really think this process should operate sample-wise, so you can use the standard meta pattern

workflows/riboseq/main.nf
```diff
@@ -533,6 +535,29 @@ workflow RIBOSEQ {
...
+        ch_merged_tracks = PLASTID_MAKE_WIGGLE.out.tracks
+            .map{ meta, tracks -> [[meta.id, tracks[0], tracks[1]]] }
+            .collect()
+            .map{ sample -> [sample.collect{it[0]}, sample.collect{it[1]}, sample.collect{it[2]}]} // id, forward_bedgraph, reverse_bedgraph
+
+        QUANTIFY_INFRAME_PSITE_PLASTID(
+            ch_merged_tracks,
+            ch_gtf.map { [ [:], it ] },
+            'gene'
+        )
```
pinin4fjords:
```suggestion                                               
      QUANTIFY_INFRAME_PSITE_PLASTID(
          PLASTID_MAKE_WIGGLE.out.tracks,        // already per-sample tuples
          ch_gtf.map { [[:], it] },                                                                                                                    
          'gene'                                                                                                                                       
      )    
```

When the process is fixed, suggest we run it per sample...

workflows/riboseq/main.nf
```diff
@@ -533,6 +535,29 @@ workflow RIBOSEQ {
...
+        REPLACE_RIBOSEQ_COUNTS_IN_MATRIX(
+            QUANTIFY_INFRAME_PSITE_PLASTID.out.counts
+                .combine(QUANTIFY_STAR_SALMON.out.counts_gene_length_scaled.map{ meta, counts -> counts })
+                .map { meta, psite_counts, salmon_counts -> [meta, [psite_counts, salmon_counts]] },
+            file("${projectDir}/bin/replace_riboseq_counts_in_matrix.awk"),
+            false
+        )
```
pinin4fjords, line 569:
```suggestion
        // Merge per-sample count files, then replace in matrix                                                                                          
        ch_psite_counts = QUANTIFY_INFRAME_PSITE_PLASTID.out.counts
            .map { meta, counts -> counts }                                                                                                              
            .collectFile(name: 'gene_inframe_psite_counts.tsv')                                                                                          
            .map { [[:], it] }

        REPLACE_RIBOSEQ_COUNTS_IN_MATRIX(                                                                                                                
            ch_psite_counts
                .combine(QUANTIFY_STAR_SALMON.out.counts_gene_length_scaled.map { meta, counts -> counts })                                              
                .map { meta, psite_counts, salmon_counts -> [meta, [psite_counts, salmon_counts]] },                                                     
            file("${projectDir}/bin/replace_riboseq_counts_in_matrix.awk"),                                                                              
            false                                                                                                                                        
        )   
```

Then something like this, leveraging collectFile()

modules/local/quantify_inframe_psite_plastid/main.nf
```diff
@@ -0,0 +1,66 @@
...
+    while IFS=\$'\\t' read SAMPLE FORWARD_BEDGRAPH REVERSE_BEDGRAPH; do
+
+        # append strand to bedgraph files and merge forward/reverse bedgraph files into a single stream
+        (
+            sed 's/\$/\\t0\\t+/' "\$FORWARD_BEDGRAPH"
+            sed 's/\$/\\t0\\t-/' "\$REVERSE_BEDGRAPH"
+        ) |
+
+        # aggregate in-frame p-site counts by gene/transcript
+        bedtools intersect -wa -wb -loj -a psites.bed -b /dev/stdin -s |
```
pinin4fjords: `-loj` generates a line for every CDS position, even with zero coverage. For a whole-genome GTF, psites.bed can have millions of rows, and left  outer join means every one of them appears in the output regardless of P-site overlap. A plain intersect (without -loj) plus adding zero counts for missing genes afterward would reduce intermediate data size significantly for large genomes.

bin/replace_riboseq_counts_in_matrix.awk
```diff
@@ -0,0 +1,33 @@
...
+#   sample \t feature \t count
+#
+# The other input file is the existing count matrix (with a header row).
+# For every sample found in the p-site counts, the corresponding column
+# in the matrix is replaced.
+
+BEGIN { FS = OFS = "\t" }
+
+# P-site counts file
+FILENAME ~ /inframe_psite_counts/ {
```
pinin4fjords: Fragile file matching. The awk script distinguishes its two input files by matching the filename against the pattern inframe_psite_counts. If the upstream process output naming ever changes, this silently breaks. Consider using FNR==NR (first-file-vs-second-file) or ARGIND instead which are order-based and don't depend on naming conventions.

nextflow_schema.json
```diff
@@ -562,9 +562,9 @@
...
-                    "help_text": "Choose 'alignment' for STAR alignment followed by Salmon in alignment-based mode (default). Choose 'pseudo' for direct Salmon pseudo-alignment, an experimental alternative that uses the same k-mer index for both modalities and may help reduce length-related quantification biases without requiring read trimming.",
+                    "enum": ["alignment", "pseudo", "plastid_psite"],
+                    "description": "Quantification method for translational efficiency analysis: 'alignment' (STAR → Salmon), 'pseudo' (Salmon pseudo-alignment) or in-frame psites (STAR → Salmon/plastid).",
+                    "help_text": "Choose 'alignment' for STAR alignment followed by Salmon in alignment-based mode (default). Choose 'pseudo' for direct Salmon pseudo-alignment, an experimental alternative that uses the same k-mer index for both modalities and may help reduce length-related quantification biases without requiring read trimming. Choose 'plastid_psite' to quantify only Ribo-seq reads with in-frame p-sites as computed by plastid (experimental); counts are scaled to the number of p-sites in the coding segment.",
```
pinin4fjords: Does this need updating now for the raw counts thing?

workflows/riboseq/main.nf
```diff
@@ -533,6 +535,29 @@ workflow RIBOSEQ {
         ch_te_counts = QUANTIFY_PSEUDO_TE.out.counts_gene_length_scaled
     }
 
+    if (params.te_quantification_method == 'plastid_psite') {
```
pinin4fjords: There should also be a `skip_plastid` check here, or there is scope for errors.

modules/local/quantify_inframe_psite_plastid/main.nf
```diff
@@ -0,0 +1,66 @@
+// Count Ribo-seq p-sites which are in-frame.
+// P-site counts are passed as bedGraph files (forward & reverse), e.g.,
+// as generated by plastid make_wiggle. The counts are summed over all
+// in-frame positions of CDS segments in a GTF file.
+
+process QUANTIFY_INFRAME_PSITE_PLASTID {
```
pinin4fjords, line 6: We also need a main.nf.test for this process

pinin4fjords (commented):
Few more things

pinin4fjords (changes requested):
Made a few comments about issues here and there, main thing is scaling problems that might be created by the looping within the new module. That should be rewired to allow nextflow to handle it.

The key changes:

- Process takes a single tuple val(meta), path(fwd), path(rev) - standard nf-core pattern
- Nextflow handles parallelism across samples instead of a shell while loop                                                                      
- collectFile merges the per-sample TSVs (3-column: sample, gene, count) into one file for the matrix replacement step
- The ugly .collect().map{ transpose } channel gymnastics in the current wiring disappear entirely since PLASTID_MAKE_WIGGLE.out.tracks already  emits per-sample tuples                                                                                                                          
                                                                                                                                                   
The GTF-to-psites.bed extraction runs redundantly per sample, so that could be factored out at this point, but it's a trivial awk operation on the GTF.

suhrig:
Thank you for the detailed feedback @pinin4fjords as always! I will address your suggestions in the upcoming days.

suhrig:
@nf-core-bot fix linting

suhrig:
@pinin4fjords I have implemented all of your requested changes, including factoring out the extraction of in-frame p-sites from the GTF file. Kindly let me know what you think of the changes.

One of the checks fails reproducibly for reasons that seem unrelated to my changes as far as I can tell. Namely, the module check for the DESEQ2_DELTA module [runs out of disk space when pulling a docker image](https://github.com/nf-core/riboseq/actions/runs/24208629560/job/70896424095?pr=141#step:4:1054). Not sure what would be the best way to fix this.

pinin4fjords:
I added https://github.com/nf-core/riboseq/pull/153 to (hopefully) fix the CI issues

pinin4fjords (approved):
Good to go!

bin/gtf_to_inframe_psites.awk
```diff
@@ -0,0 +1,29 @@
...
+BEGIN { FS = OFS = "\t" }
+
+$3 == "CDS" && $9 ~ (FEATURE "_id \"") {
+
+    # parse GTF fields
+    chrom = $1; start = $4; end = $5; strand = $7; phase = $8
+    feature = $9; sub(".*" FEATURE "_id \"", "", feature); sub(/".*/, "", feature)
+
+    # print deduplicated coordinates of codon starts
+    frame = strand == "+" ? phase : (3 - phase + end - start) % 3
```
JackCurragh, line 24: This frame calculation assumes the GFF3 definition of phase but the file and params refer to GTF. Might be confusing. 

GFF3 phase says how many bases to skip while Ensembl GTFs say bases from the previous exon completing this codon. AT -splice- G would be phase 1 for GFF3 but 2 for GTF. 

Nightmare, I know.

workflows/riboseq/main.nf
```diff
@@ -533,6 +536,41 @@ workflow RIBOSEQ {
         ch_te_counts = QUANTIFY_PSEUDO_TE.out.counts_gene_length_scaled
     }
 
+    if (params.te_quantification_method == 'plastid_psite' && !params.skip_plastid) {
+        // Convert GTF CDS segments to in-frame p-site positions
+        GTF_TO_INFRAME_PSITES(
+            ch_gtf.map { gtf -> [ [id: gtf.baseName, feature: 'gene'], gtf ] },
```
JackCurragh, line 542: Intentionally hardcoded to gene features?

JackCurragh (changes requested):
I think the in-frame psites awk is the only thing needing a fix. I think either moving to GFF usage or converting to GTF logic as is stated in the file

bin/gtf_to_inframe_psites.awk
suhrig: Thanks for taking a thorough look. Are you sure about this, though? The format specs describe both fields to be interpreted the same:

GTF: https://www.ensembl.org/info/website/upload/gff.html
GFF3: https://www.ensembl.org/info/website/upload/gff3.html

I also checked GENCODE and ENSEMBL empirically and in both cases the phases are the same for the GTF and GFF3 file (apart from a very very small percentage, which I attribute to formatting or other negligible factors):

```
# download GENCODE GTF/GFF3
wget https://ftp.ebi.ac.uk/pub/databases/gencode/Gencode_human/release_49/gencode.v49.annotation.gtf.gz
wget https://ftp.ebi.ac.uk/pub/databases/gencode/Gencode_human/release_49/gencode.v49.annotation.gff3.gz

# compare phases
diff \
    <(zcat gencode.v19.annotation.gff3.gz | awk '$3=="CDS"' | sort -k1,1V -k4,4n -k5,5n | cut -f8) \
    <(zcat gencode.v49.annotation.gtf.gz  | awk '$3=="CDS"' | sort -k1,1V -k4,4n -k5,5n | cut -f8) \
| wc -l

# result: 762

# download ENSEMBL GTF/GFF3
wget https://ftp.ensembl.org/pub/release-115/gtf/homo_sapiens/Homo_sapiens.GRCh38.115.gtf.gz
wget https://ftp.ensembl.org/pub/release-115/gff3/homo_sapiens/Homo_sapiens.GRCh38.115.gff3.gz

# compare phases
diff \
    <(zcat Homo_sapiens.GRCh38.115.gff3.gz | awk '$3=="CDS"' | sort -k1,1V -k4,4n -k5,5n | cut -f8) \
    <(zcat Homo_sapiens.GRCh38.115.gtf.gz  | awk '$3=="CDS"' | sort -k1,1V -k4,4n -k5,5n | cut -f8) \
| wc -l

# result: 2498
```

workflows/riboseq/main.nf
suhrig: Currently, yes. This should be parameterized when the pipeline supports transcript-level processing. But that is currently not the case, so no need to provide a parameter that accepts only one valid option.

bin/gtf_to_inframe_psites.awk
JackCurragh: Interesting... I guess I was mistaken. I did a similar check against refseq just now and all seems consistent there too.

JackCurragh (approved):

PR #142, opened 2026-01-19
Add ribodetector_chunk_size parameter for memory management

pinin4fjords:
## Summary
- Adds `ribodetector_chunk_size` parameter (default: 100) to control RiboDetector memory usage
- Prevents extreme memory consumption that can occur with large datasets when chunk_size is not set
- Parameter is conditionally passed to RiboDetector only when set

## Context
RiboDetector without `--chunk_size` can consume excessive memory (reported cases of 36GB+ with continued failures even at 688GB). Setting `--chunk_size 100` keeps memory linear to thread count without affecting throughput.

## Test plan
- [ ] Run pipeline with `--ribo_removal_tool ribodetector` and verify chunk_size is passed
- [ ] Verify memory usage is manageable with the default setting


suhrig (approved):
All good.

Speaking of resource consumption: Is there any particular reason that the number of threads is fixed to 1? Why not more?

pinin4fjords:
> All good.
> 
> Speaking of resource consumption: Is there any particular reason that the number of threads is fixed to 1? Why not more?

good point. Probably just a testing hangover, I'll fix

FelixKrueger (approved):
LGTM.

PR #143, opened 2026-01-19
Remove pipeline-level fixed seeds from production config

pinin4fjords:
## Summary

- remove pipeline-supplied fixed seeds from fq subsampling and both STAR alignment paths
- set explicit seeds in the nf-test config for fq, STAR, and UMI-tools snapshot stability
- move the UMI-tools wrapper policy upstream in [nf-core/modules#12662](https://github.com/nf-core/modules/pull/12662)

## Behaviour

Omitting the fq seed makes fresh production runs initialise from entropy, so subsampling variability is visible rather than silently fixed.

STAR is different: its upstream default remains `runRNGseed 777`, and the pipeline currently uses `outMultimapperOrder Old_2.4` rather than random multimapper ordering. Removing the pipeline override therefore delegates to STAR and removes a pipeline-level policy; it does not claim STAR becomes stochastic between runs.

UMI-tools randomly chooses a representative read when tied duplicate candidates have equal mapping quality. The upstream module PR removes its injected CLI seed and `PYTHONHASHSEED`, while retaining explicit seeds in test configuration.

## Pipeline audit

- RiboDetector is already unseeded in production and seeded in tests.
- Bowtie2 has an implicit default seed, but the pipeline does not supply one; enabling `--non-deterministic` is a separate algorithmic-mode change.
- kallisto uses its seed for bootstrap sampling, while this pipeline defaults to zero bootstraps.

No production changes are needed for those tools.

## Tests

- targeted `prek` hooks passed
- `nf-core pipelines lint` passed
- x86_64 Docker: both focused UMI pipeline tests passed, including the extended-ORF path


FelixKrueger (commented):
Two things to flag here, on top of the merge conflict with `dev`. The STAR half may not do what it's aiming for: STAR's own default is `runRNGseed 777`, so dropping `--runRNGseed 0` swaps one fixed seed for another rather than removing determinism. And `dev` has since grown a **second** `--runRNGseed 0`, on the extended-ORF second-pass hybrid alignment, which this branch predates — so merging as-is would leave the two STAR passes running on different seeds, which seems worse than either uniform choice. The `FQ_SUBSAMPLE` half does work as intended: `fq` initialises its RNG from entropy when `--seed` is omitted, so that one genuinely makes subsampling vary run to run. No objection from me either way, but it's a real behaviour change rather than a config tidy-up, so worth being deliberate about. What would you like to do about the second STAR site?

<details>
<summary>Evidence (AI-assisted)</summary>

**STAR's default is also fixed** — `alexdobin/STAR`, `source/parametersDefault`:

```
runRNGseed                      777
    int: random number generator seed.
```

So after this change STAR is still deterministic, just seeded 777 instead of 0. Whatever multimapper tie-breaking the seed drives will shift, without the PR's stated goal being reached.

**`fq` randomises when the flag is absent** — `stjude-rust-labs/fq`, `src/commands/subsample.rs`:

```rust
let rng = if let Some(seed) = args.seed {
    info!(seed = seed, "initializing rng from seed");
    SmallRng::seed_from_u64(seed)
} else {
    info!("initializing rng from entropy");
    rand::make_rng()
};
```

**Two seed sites on current `dev`**, where this branch has none:

| `conf/modules.config` | Selector |
|---|---|
| line 585 | `withName: 'STAR_ALIGN'` |
| line 1406 | `withName: '.*:EXTENDED_ORF_SECOND_PASS_ALIGN:FASTQ_ALIGN_STAR_HYBRID:STAR_ALIGN'` |

The branch removed the only occurrence that existed in January; line 1406 arrived with the second-pass work afterwards.

**State:** `mergeStateStatus: DIRTY` / `mergeable: CONFLICTING`, so it needs a merge from `dev` regardless. The checks currently shown are from 2026-01-19 and will re-run once it is updated.

</details>

pinin4fjords:
@FelixKrueger let's put this one on hold, I think the community needs some consistency- discussions ongoing (e.g. https://nfcore.slack.com/archives/CQY2U5QU9/p1786549975978849)

PR #145, opened 2026-02-23
Filter to primary alignments before UMI-tools dedup

pinin4fjords:
## Summary

UMI-tools dedup processes all alignments in the input BAM, including secondary and supplementary alignments from multi-mapping reads. This inflates the dedup statistics, making them disconnected from the rest of the QC flow.

For example, from a real run: a sample with ~26M primary alignments (from ~36M total mapped reads including multi-mappers, ~73M total alignment records) reported 73M input reads and 12M output in the UMI-tools log. The actual primary-only dedup numbers were 26M in / 7.9M out, giving a true duplication rate of ~70% rather than the misleading ~84% implied by the raw UMI-tools stats.

This makes it impossible to track reads through the QC flow: raw reads > trimmed > non-rRNA > aligned > deduplicated.

### What this PR does

- Adds a `samtools view -F 0x900` filtering step before UMI-tools dedup to remove secondary (0x100) and supplementary (0x200) alignments
- Uses the existing nf-core `samtools/view` module (aliased as `SAMTOOLS_VIEW_PRIMARY`) so no custom containers are needed
- Controlled by `--umitools_dedup_primary_only` (default: `true`)
- The filtered BAM is an intermediate (not published) - the original unfiltered BAM with all multi-mapping information is still available in the sorted alignment output
- Only applies to the `umitools` dedup tool path (not `umicollapse`)
- Subworkflows updated from nf-core/modules via `nf-core subworkflows update` after upstream PR was merged: https://github.com/nf-core/modules/pull/10135

### Topic-based version collection

Several of the updated nf-core modules (e.g. `samtools/view`, `samtools/index`) now emit software versions via Nextflow topic channels rather than traditional `versions.yml` files. Without handling these, the pipeline's version report would be missing entries for those tools.

This PR adds topic version collection to the pipeline (following the pattern from nf-core/rnaseq):

- Subscribes to the `versions` topic channel and branches entries into `Path` objects (traditional `versions.yml` files) and tuples (topic-based `[process, tool, version]`)
- Converts topic tuples into YAML-formatted version strings, deduplicating and sorting by process
- Mixes both sources into `softwareVersionsToYAML()` so all tool versions appear in the final MultiQC report

### Approaches considered

Several alternatives were discussed:

1. **Filter to primary alignments before dedup** (this PR) - fixes the stats at source by ensuring UMI-tools only sees primary alignments. The original BAM with all mapping locations is preserved in the sorted alignment output.

2. **Limit STAR to single alignments** (`--outFilterMultimapNmax 1`) - rejected because it discards multi-mappers entirely, which are needed for Salmon quantification, and misrepresents alignment accuracy.

3. **Manual counting with samtools** independently of UMI-tools - correct but more complex to implement and would require replacing the existing MultiQC UMI-tools module with custom content.

4. **Run UMI-tools as-is but filter the stats** - would still waste compute deduplicating secondary alignments and would need custom MultiQC integration.

Option 1 was chosen as it fixes the problem at source, reuses existing nf-core modules, preserves the native UMI-tools MultiQC integration, and keeps the original unfiltered BAM available for users who need multi-mapping information.

### Files changed

- `nextflow.config` - new `umitools_dedup_primary_only` parameter (default: `true`)
- `nextflow_schema.json` - schema entry in UMI options group
- `conf/modules.config` - process config for `SAMTOOLS_VIEW_PRIMARY` (`-F 0x900 -b`) and `SAMTOOLS_INDEX_PRIMARY` (both unpublished)
- `workflows/riboseq/main.nf` - passes `params.umitools_dedup_primary_only` to `BAM_DEDUP_UMI`; adds topic-based version collection and merges both version sources into `softwareVersionsToYAML()`
- `subworkflows/nf-core/bam_dedup_stats_samtools_umitools/` - updated from nf-core/modules; conditional `SAMTOOLS_VIEW_PRIMARY` + `SAMTOOLS_INDEX_PRIMARY` before `UMITOOLS_DEDUP`
- `subworkflows/nf-core/bam_dedup_umi/` - updated from nf-core/modules; passes `primary_only` parameter to both genome and transcriptome dedup calls
- `modules.json` - updated component SHAs

## Test plan

- [ ] Stub run passes: `nextflow run . -profile test,docker --with_umi --umitools_dedup_primary_only --outdir results -stub`
- [ ] Full test run with UMI data confirms dedup stats now reflect primary alignment counts
- [ ] Verify `--umitools_dedup_primary_only false` preserves the old behaviour
- [ ] Verify new parameter appears in `nextflow run . --help`
- [ ] Verify software versions report includes tools emitting via topic channels (e.g. samtools/view)


maxulysse (approved):

iraiosub:
I am happy with this solution - approved. Two questions: 

- it seems to me this is sth to be considered to be fixed upstream i.e. in UMItools - maybe if an issue is raised, they'll pick it up?  
- is this behaviour affecting other workflows? rnaseq + any others using the UMItools module?

iraiosub (approved):

pinin4fjords:
> I am happy with this solution - approved. Two questions:
> 
> * it seems to me this is sth to be considered to be fixed upstream i.e. in UMItools - maybe if an issue is raised, they'll pick it up?

I'm pretty sure there are issues, unresolved. I'm not sure how maintained UMItools is right now tbh

> * is this behaviour affecting other workflows? rnaseq + any others using the UMItools module?

Yes, once we've tested it a bit here, we can port it. The suborkflow is already updated in nf-core/modules, so that will be trivial.

PR #147, opened 2026-03-02
Fix RiboCode error handling and P-site detection failures

pinin4fjords:
## Summary

- **Update ribocode modules** from nf-core/modules with error handling fixes:
  - `ribocode/ribocode`: Replace `2>&1 || test -s` with `tee` + grep error check, removing `|| true` that masked OOM kills from Platform (nf-core/modules#9581, nf-core/modules#10295)
  - `ribocode/metaplots`: Add config validation - fail early with a clear error when no periodic signal is detected, instead of passing empty config to RiboCode
  - Add failure tests for both modules
- **Relax metaplots thresholds in test_full** (`-pv1 0.05 -pv2 0.05 -f0_percent 0.5`) so RiboCode runs successfully on the full test dataset
- **Add troubleshooting docs** for the P-site detection failure, documenting the `extra_ribocode_metaplots_args` workaround

## Test plan

- [ ] Run test_full - RiboCode should complete with relaxed thresholds
- [ ] Confirm Platform correctly reports failures (no more masked exit codes)


JackCurragh (approved):
This looks good to me. Informative rather than over complicating it.

PR #151, opened 2026-03-06
fix: always skip igenomes_base validation

pinin4fjords:
## Summary

- Always ignore `igenomes_base` during parameter validation
- Add the fix to the changelog

nf-schema 2.5.0+ validates cloud storage paths used by path-formatted parameters. Users without AWS credentials can otherwise get a 403 when validation checks the default S3 iGenomes path. `igenomes_base` is only used to derive individual reference paths, and those paths are validated separately, so the base path itself should be ignored.

This follows the recommendation in [nextflow-io/nf-schema#204](https://github.com/nextflow-io/nf-schema/issues/204). The nf-core 4.0.2 template sync already removed `format: directory-path` from the riboseq schema on `dev`; this PR adds the remaining pipeline-level ignore, matching [nf-core/rnaseq#1745](https://github.com/nf-core/rnaseq/pull/1745) and the current rnaseq `dev` configuration.

## Test plan

- [x] Verify the pipeline validates other parameters normally
- [x] Confirm validation without AWS credentials does not access `igenomes_base`
- [x] Confirm an explicit `igenomes_base` derives the expected reference paths and those paths are validated
- [x] Run `nf-core pipelines lint`


JackCurragh (approved):
Looks good to me

PR #154, opened 2026-04-30
feat: integrate riboWaltz MultiQC module (1.34) and align FastQC/SortMeRNA general-stats handling with rnaseq

pinin4fjords:
## Summary

Integrates the riboWaltz MultiQC module ([MultiQC#3465](https://github.com/MultiQC/MultiQC/pull/3465), shipped in MultiQC 1.34) so riboWaltz QC plots and stats land in the pipeline's MultiQC report alongside everything else, replacing the bespoke custom-content shim that lived on a previous closed PR (#139). While validating the merged general-stats table, also brings the FastQC + SortMeRNA setup into line with `nf-core/rnaseq` so the two pipelines stay easy to maintain together.

End result: P-site region distribution, reading-frame distribution, start/stop-codon metaprofiles, and `cds_pct` / `frame0_cds_pct` general-stats columns appear in the existing MultiQC report, merged onto the same row as each sample's other QC metrics. Paired-end FastQC values now populate correctly across raw / trimmed / filtered stages, with column titles disambiguated.

## Changes

**riboWaltz integration**

- Bump `nf-core/multiqc` module 1.33 → 1.34 (the version that ships the riboWaltz plugin).
- `ext.prefix = "${meta.id}.ribowaltz"` on `RIBOWALTZ` so output TSVs are named `<sample>.ribowaltz.{psite_region,frames,metaprofile_psite}.tsv`. The plugin's default search pattern is `*ribowaltz*…`, which is a fast filename pre-filter; matching the convention here keeps file scanning cheap.
- Mix `RIBOWALTZ.out.ribowaltz_qc_data` into `ch_multiqc_files`.
- `assets/multiqc_config.yml`: add `ribowaltz` to `run_modules` and `module_order`; `sample_names_replace_regex` strips the `.ribowaltz` suffix off the in-file `sample` column inside the QC TSVs so general-stats rows merge with each sample's other-tool entries.

**FastQC / SortMeRNA general-stats handling (aligned with rnaseq)**

- `use_filename_as_sample_name: [sortmerna]` — SortMeRNA logs two `Reads file:` lines for paired-end data and the parser picks R2, giving every sample a spurious `_2` suffix that broke `table_sample_merge` grouping.
[...]

PR #155, opened 2026-04-30
fix: drop conditional wrappers in conf/modules.config for Nextflow 26.04

pinin4fjords:
## Summary

Nextflow 26.04.0 (released 2026-04-30) enabled the v2 config parser by default. The v2 parser is strictly declarative and rejects mixing `if (...)` control flow with config statements — `'If statements cannot be mixed with config statements'`. As a result `nextflow config -flat .` fails to parse this pipeline's config, which breaks `nf-core pipelines lint` on every PR (e.g. #154).

## Changes

Unwrap the 19 top-level (and a few nested) `if (params.skip_X) { process { withName: ... { ... } } }` blocks in `conf/modules.config`.

After the change:

- `withName` directives are always declared.
- The actual gating already happens in `workflows/riboseq/main.nf`, where each conditionally-invoked process is guarded (e.g. `if (!params.skip_bbsplit) BBMAP_BBSPLIT(...)`).
- The v2 parser's `ConfigSpecVisitor` (mirrored from the Nextflow language server) explicitly excludes process selectors from scope validation, so always-defined `withName` blocks for processes the workflow may skip don't trigger `'unrecognized config option'` warnings.
- This matches the pattern nf-core/rnaseq has used for some time.

## Verification

`nextflow config -flat .` on Nextflow 26.04.0 now parses cleanly. The visible diff is the 19 unwraps + dedent; no behavioural changes.

## Test plan

- [x] `nextflow config -flat .` parses on Nextflow 26.04.0
- [ ] CI lint job passes
- [ ] CI nf-test matrix (25.04.8 and latest-everything) passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#145](https://github.com/nf-core/riboseq/pull/145) - Filter to primary alignments before UMI-tools dedup ([@pinin4fjords](https://github.com/pinin4fjords))
 - [#147](https://github.com/nf-core/riboseq/pull/147) - Fix RiboCode error handling and P-site detection failures ([@pinin4fjords](https://github.com/pinin4fjords))
 - [#148](https://github.com/nf-core/riboseq/pull/148) - Update all nf-core modules/subworkflows to latest versions, migrate to topic-based version reporting, fix publishing paths for new subworkflow nesting ([@pinin4fjords](https://github.com/pinin4fjords))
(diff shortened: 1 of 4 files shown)


JackCurragh (approved):
All looks good to me. 

I would be interested to know why it is preferable to use the `def strand_filter` formulation over what was there previously though

pinin4fjords:
> I would be interested to know why it is preferable to use the `def strand_filter` formulation over what was there previously though

Nextflow 26.04's v2 config/script parser doesn't accept a map key/value pair split across a line break — the original `strand_filter:\n        getStrandFilter(...)` blows up with `Unexpected input: '\n'` right at the colon's position. Keeping the key and its value on the same line clears it; lifting `getStrandFilter(...)` into a local was just to stop the line getting unreadable. Could equally be one long inline call.

PR #158, opened 2026-05-01
feat(stringtie): add reference-guided novel transcript discovery

### LINKED ISSUE #157: Port StringTie-based novel transcript discovery from nf-core/rnaseq
### Motivation

riboseq currently relies entirely on the user-supplied GTF for transcript definitions. That limits utility for studies of unannotated or non-canonical ORFs (micro-proteins, novel uORFs/dORFs etc.) where the annotation is the bottleneck. The ask is for an optional transcript-detection mode that discovers novel transcripts from the alignments and merges them with the reference annotation before downstream ORF / P-site analysis.

nf-core/rnaseq already does this with StringTie reference-guided assembly plus StringTie merge. Porting that machinery here gets us most of the way to the novel-transcript requirement without reinventing it.

### What rnaseq has

- `subworkflows/nf-core/bam_stringtie_merge` — per-sample reference-guided assembly via `STRINGTIE_STRINGTIE`, then `STRINGTIE_MERGE` collapses per-sample GTFs into a unified annotation.
- A final `STRINGTIE_STRINGTIE` quantification step run against either the merged or the reference GTF.
- Params: `skip_stringtie` (off by default in rnaseq) and `stringtie_ignore_gtf` (toggles guided vs purely de novo assembly).
- Module config at `conf/modules/stringtie.config` with publish layout for the per-sample, merged, and post-merge quantification steps. rnaseq runs the StringTie modules at tool defaults (only adding `-v` for verbosity); no abundance/coverage overrides.

Recent fixes worth carrying over:

- nf-core/rnaseq@9a0835e — avoids a publish collision between the outer `STRINGTIE_STRINGTIE` and the one inside `BAM_STRINGTIE_MERGE`, publishes the merged GTF, documents the de novo flow.
- nf-core/rnaseq@7bad7ef — cleans up `BAM_STRINGTIE_MERGE` wiring.

### Proposed approach

1. Install `bam_stringtie_merge` via `nf-core subworkflows install` — that pulls in `stringtie/stringtie` and `stringtie/merge` automatically.
2. Wire it after `FASTQ_ALIGN_STAR`, taking `ch_genome_bam` and `ch_gtf`.
3. Add params: `skip_stringtie` (default `true`, see below), `stringtie_ignore_gtf` (default `false`), plus merge-stage knobs exposed for users who want to tighten: `stringtie_merge_min_tpm` (maps to merge `-T`, default `0` per StringTie), `stringtie_merge_min_iso_frac` (maps to merge `-f`, default `0.01`), `stringtie_keep_retained_introns` (maps to merge `-i`, default `false`).
4. Route the merged GTF emitted by `bam_stringtie_merge` into the downstream channels currently driven from `ch_gtf` — at minimum the transcriptome FASTA build that feeds Salmon, RiboCode, and riboWaltz.

### Filtering strategy

Run StringTie at tool defaults (per-sample `-c 1 -j 1 -f 0.01 -m 200`, merge `-T 0 -F 0 -f 0.01 -m 50` without `-i`), matching rnaseq's precedent. Don't layer Ribo-seq specific filtering (3-nt periodicity, P-site coverage, ORF length, start-codon scoring) onto the GTF: that work belongs in the ORF caller. The Ribo-seq smORF tool benchmark by [Tong, Hah & Martinez (Brief Bioinform 2024)](https://doi.org/10.1093/bib/bbae268) shows that all five major Ribo-seq ORF callers (RibORF v0.1/v1.0, RiboCode, ORFquant, Ribo-TISH) take a pre-assembled transcriptome as input and apply their filtering at the ORF level — periodicity p-values, in-frame fraction, length cutoffs. Filtering at the transcript stage would discard candidates before the ORF callers ever see them.

For users who do want to tighten the merge, the params above expose the relevant StringTie knobs without forcing a non-default choice on the pipeline as a whole. As points of reference: the [RIBOSS pipeline (Lim et al., Brief Bioinform 2025)](https://doi.org/10.1093/bib/bbaf164) sets per-sample `-f 0.1` for stricter isoform filtering, while [Wu et al. (Plant Physiol 2019)](https://doi.org/10.1104/pp.19.00541) keep all StringTie class codes and apply only a post-hoc TPM > 0.5 cut for downstream stats — the right setting depends on the organism and the downstream question.

### Default: `skip_stringtie = true`

Inverting the rnaseq default. Most riboseq users run against well-annotated references (human, mouse, yeast, zebrafish) where a guided StringTie assembly will, after merge, be near-identical to the supplied GTF — at the cost of an extra StringTie pass per sample, the merge step, and a forced Salmon re-index against the merged transcriptome. Unlike rnaseq, where StringTie output is a side product, here the merged GTF actively rewires Salmon, RiboCode, and riboWaltz inputs, so it is not a free add-on. Novel-ORF / micro-protein discovery is a deliberate, power-user workflow; users who want it will read the docs and flip the flag.

### Refs

- StringTie manual: https://ccb.jhu.edu/software/stringtie/index.shtml?t=manual
- nf-core/rnaseq subworkflow: https://github.com/nf-core/rnaseq/tree/dev/subworkflows/nf-core/bam_stringtie_merge
- nf-core/rnaseq stringtie config: https://github.com/nf-core/rnaseq/blob/dev/conf/modules/stringtie.config

pinin4fjords:
Closes #157.

## Summary

- Install `nf-core/bam_stringtie_merge` subworkflow plus `stringtie/stringtie` (2.2.3) and `stringtie/merge` (2.2.1) modules.
- Add `--skip_stringtie` (default `true` - inverted vs. rnaseq, novel-ORF discovery is a power-user workflow), `--extra_stringtie_args`, `--extra_stringtie_merge_args`.
- Publish the merged GTF as a side product under `<outdir>/stringtie/stringtie_merge.gtf` plus per-sample `<id>.denovo.transcripts.gtf` assemblies.
- The rest of the pipeline continues to consume the supplied reference GTF. **The merged GTF is not yet wired back into Ribo-TISH / Ribotricer / plastid / GTF_TO_INFRAME_PSITES** - see "Decisions" below.

## Required follow-on nf-core/modules changes

Wiring the merged GTF into the riboseq ORF callers needs upstream module updates first. Concrete asks (file as separate PRs against [nf-core/modules](https://github.com/nf-core/modules)):

- [ ] **`ribotish/predict`** - add an optional `tuple val(meta), path(alt_gtf)` input that maps to `-a` on the CLI. This is the classification annotation Ribo-TISH uses to tag ORFs as `uORF` / `dORF` / `annotated` / `CDSFrameOverlap` / `Novel`. Without it, every ORF found on the merged GTF would be tagged generically.
- [ ] **`ribotish/quality`** - same shape (optional secondary GTF -> `-a`). Less critical (quality currently runs on reference and stays there), but worth aligning so a future "all-merged-GTF" mode is consistent.
- [ ] **`ribotricer/prepareorfs`** - investigate whether the upstream Ribotricer CLI has an equivalent secondary-annotation flag for ORF classification (`prepare-orfs --annotation`?). If yes, add an optional input. If no, file an upstream Ribotricer feature request.

Once those modules expose secondary annotation, the riboseq-side follow-on PR is small: bump the modules, change the affected call sites in `workflows/riboseq/main.nf` to pass `merged_gtf` as primary and `reference_gtf` as secondary, and the merged GTF starts informing novel-ORF discovery.

Plastid and `GTF_TO_INFRAME_PSITES` stay on the reference GTF regardless - they're CDS-bound for ROI definition / read filtering and there isn't a sensible "secondary annotation" semantics for them.

## Decisions to revisit

These were debated in this PR's iteration; capturing them so we can revisit if context shifts.

### 1. Merged GTF is published only, not consumed by ORF callers - for now

**Why:** the riboseq ORF callers split into two correctness regimes:

- CDS-aware tools (Ribo-TISH `quality`, plastid metagene, `GTF_TO_INFRAME_PSITES`) need reference CDS for read filtering or ROI definition. StringTie's merged GTF only emits `transcript` and `exon` features, no `CDS`. Feeding the merged GTF here gives `Counted reads: 0. Error: no reads found! Check protein coding annotation.` (verified on the VM at `0c5e02bf` before this revert).
- De novo ORF scanners (Ribo-TISH `predict`, Ribotricer `prepare-orfs`, RiboCode) **can** scan a CDS-less GTF for ORFs - that's how they work. But to *classify* discovered ORFs (uORF / dORF / annotated / `CDSFrameOverlap` / Novel) they need the reference CDS as a secondary annotation. Ribo-TISH's CLI exposes `-a SECONDARY` for exactly this case.

**Blocker:** the nf-core modules `ribotish/quality`, `ribotish/predict`, and `ribotricer/prepareorfs` all take a single GTF input - no slot for a secondary annotation. See the "Required follow-on" checklist above.

**Plan:** land this PR as side-product-only; the follow-on lands the proper wiring once the modules are updated.

### 2. Transcriptome-BAM-bound tools are intentionally excluded

RiboCode, riboWaltz, and alignment-mode Salmon all consume STAR's transcriptome BAM, which was keyed to the reference GTF at alignment time. Swapping in a different GTF post-hoc would create a transcript-ID mismatch. A future opt-in (`stringtie_realign` or similar) could rebuild the Salmon index and re-run STAR with the merged GTF; out of scope for this PR.

### 3. TransDecoder pre-pass for sequence-level CDS prediction on novel transcripts

Considered and rejected for now. The Ribo-seq ORF callers do their own ORF scanning + filtering on transcripts using Ribo-seq evidence (triplet periodicity, P-site coverage). A TransDecoder pre-pass would add a sequence-level filter (Markov ORF score, BLAST, Pfam) that's biased differently - it could discard short or unusual ORFs that Ribo-seq evidence would have validated. The Ribo-seq tools' built-in scanners are the right discovery layer; TransDecoder would only matter as a side artefact for proteomics/orthogonal validation.

If we want sequence-level ORF prediction as a separate output, it should be a distinct opt-in flag with its own deliverable, not coupled to the main StringTie wiring.

### 4. `-e` is intentionally absent from the per-sample StringTie call

`-e` restricts StringTie's output to transcripts already in the reference GTF, which would silently make the merged GTF a clone of the reference and turn the whole feature into a no-op. Mirrors the fix in nf-core/rnaseq commit `9a0835e`.

### 5. CLI ergonomics caveat

`--skip_stringtie false` from the CLI is rejected by the nf-schema plugin as a string when the schema default is `true`. Users need to override via `-c <config>` or `-params-file`. This is an nf-schema behaviour (also affects other `default: true` boolean params in the pipeline), not specific to this PR.

## Implementation notes

- The per-sample StringTie call inside the subworkflow uses `-v` only.
- The subworkflow-internal `STRINGTIE_STRINGTIE` is matched by a process selector scoped to `NFCORE_RIBOSEQ:RIBOSEQ:BAM_STRINGTIE_MERGE:STRINGTIE_STRINGTIE` and emits `<id>.denovo.transcripts.gtf`. Per-sample abundance/coverage outputs are not published (not comparable across samples pre-merge).
- Schema entries split: `skip_stringtie` is in the skip-* group; `extra_stringtie_*_args` are in the extra-args group with the rest of the per-tool extras.

## Test plan

- [x] `nf-core pipelines schema lint` clean.
- [x] `nf-core pipelines lint` - 0 failures.
- [x] `nextflow inspect` resolves `STRINGTIE_STRINGTIE` and `STRINGTIE_MERGE` containers when StringTie is on.
- [x] **VM end-to-end with `--skip_stringtie false`**: pipeline completed in 10m 17s, 359 tasks succeeded. Per-sample `*.denovo.transcripts.gtf` published for all 12 samples; merged `stringtie_merge.gtf` published (8.7 MB, 52,032 `MSTRG` records). All downstream consumers (Ribo-TISH quality + predict, riboWaltz, plastid metagene/psite/wiggle, Salmon, anota2seq, MultiQC) ran cleanly against the unchanged reference GTF.
- [ ] CI on this push.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#141](https://github.com/nf-core/riboseq/pull/141) - Add in-frame P-site quantification method ([@suhrig](https://github.com/suhrig))
 - [#142](https://github.com/nf-core/riboseq/pull/142) - Add `ribodetector_chunk_size` parameter to control RiboDetector memory usage ([@pinin4fjords](https://github.com/pinin4fjords))
 - [#144](https://github.com/nf-core/riboseq/pull/144) - Add bigWig coverage tracks ([@suhrig](https://github.com/suhrig))
+- [#157](https://github.com/nf-core/riboseq/issues/157) - Add optional StringTie reference-guided novel transcript discovery; merged GTF feeds Ribo-TISH, Ribotricer, plastid, and the in-frame p-site quantification ([@pinin4fjords](https://github.com/pinin4fjords))
 
 ### `Fixed`
 
@@ -58,6 +59,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 |               | `--extra_plastid_make_wiggle_args`       |
 |               | `--skip_coverage_tracks`                 |
 |               | `--ribodetector_chunk_size`              |
+|               | `--skip_stringtie`                       |
+|               | `--extra_stringtie_args`                 |
+|               | `--extra_stringtie_merge_args`           |
 
 ### `Dependencies`
 
@@ -67,6 +71,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 | `plastid`          |             | 0.6.1       |
 | `bedtools`         |             | 2.31.1      |
 | `bedGraphToBigWig` |             | 469         |
+| `StringTie`        |             | 2.2.3       |
 
 ## v1.2.0 - 2025-12-03
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#141](https://github.com/nf-core/riboseq/pull/141) - Add in-frame P-site quantification method ([@suhrig](https://github.com/suhrig))
 - [#142](https://github.com/nf-core/riboseq/pull/142) - Add `ribodetector_chunk_size` parameter to control RiboDetector memory usage ([@pinin4fjords](https://github.com/pinin4fjords))
 - [#144](https://github.com/nf-core/riboseq/pull/144) - Add bigWig coverage tracks ([@suhrig](https://github.com/suhrig))
-- [#157](https://github.com/nf-core/riboseq/issues/157) - Add optional StringTie reference-guided novel transcript discovery; merged GTF feeds Ribo-TISH, Ribotricer, plastid, and the in-frame p-site quantification ([@pinin4fjords](https://github.com/pinin4fjords))
+- [#157](https://github.com/nf-core/riboseq/issues/157) - Add optional StringTie reference-guided novel transcript discovery; merged GTF published as a side product (downstream wiring deferred to a follow-on, see PR #158) ([@pinin4fjords](https://github.com/pinin4fjords))
 
 ### `Fixed`
 
--- a/conf/modules.config
+++ b/conf/modules.config
@@ -1062,4 +1062,30 @@ process {
         ]
     }
 
+    // `.denovo` prefix and the saveAs filter avoid a publish collision with any
+    // future standalone STRINGTIE_STRINGTIE call and drop the per-sample
+    // abundance/coverage outputs (not comparable across samples pre-merge).
+    // No `-e`: that flag would restrict output to reference transcripts and
+    // make novel discovery a silent no-op.
+    withName: 'NFCORE_RIBOSEQ:RIBOSEQ:BAM_STRINGTIE_MERGE:STRINGTIE_STRINGTIE' {
+        ext.args   = { ['-v', params.extra_stringtie_args ?: ''].join(' ').trim() }
+        ext.prefix = { "${meta.id}.denovo" }
+        publishDir = [
+            path: { "${params.outdir}/stringtie" },
+            mode: params.publish_dir_mode,
+            saveAs: { filename ->
+                if (filename.endsWith('.transcripts.gtf')) return filename
+                return null
+            }
+        ]
+    }
+
+    withName: 'NFCORE_RIBOSEQ:RIBOSEQ:BAM_STRINGTIE_MERGE:STRINGTIE_MERGE' {
+        ext.args   = { params.extra_stringtie_merge_args ?: '' }
+        publishDir = [
+            path: { "${params.outdir}/stringtie" },
+            mode: params.publish_dir_mode,
+            saveAs: { filename -> filename.equals('versions.yml') ? null : filename }
+        ]
+    }
 }
--- a/docs/output.md
+++ b/docs/output.md
@@ -293,6 +293,19 @@ The pipeline produces coverage tracks in bigWig format. In the case of stranded 
   - `<SAMPLE>.reverse.bigWig`: Coverage on the reverse strand (only created for stranded libraries)
   - `<SAMPLE>.unstranded.bigWig`: Sum of coverage of forward and reverse strand (only created for unstranded libraries)
 
+## Novel transcript discovery (StringTie)
+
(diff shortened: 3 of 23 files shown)


pinin4fjords:
Closing as superseded. This original StringTie foundation is being rebuilt as part of the stacked-PR split of #174: the StringTie install + hybrid GTF chain + RNA-seq BAM preference + class-code filter all land together in the new feat/164-stringtie-novel branch (PR opening shortly), wired into the canonical-GTF backbone from #161.

PR #177, opened 2026-05-22
feat: canonical annotation backbone via --canonical_gtf

pinin4fjords:
## Summary

Adds the one-transcript-per-gene canonical annotation backbone (MANE Select, Ensembl canonical, or equivalent) used by the genome-coordinate ORF callers (Ribo-TISH, Ribotricer), plastid P-site quantification, and the translational-efficiency analysis. Most Ribo-seq reads land on exons shared between isoforms and cannot be assigned to a specific transcript (~7% map uniquely with full multi-isoform annotation, Wang et al., Bioinformatics 2016); restricting to one representative transcript per gene eliminates the ambiguity by design.

## Changes

- New param `--canonical_gtf` accepts a user-supplied canonical GTF (MANE Select for human, Ensembl `canonical` tag for other Ensembl species).
- When `--canonical_gtf` is not supplied, `AGAT_SP_KEEP_LONGEST_ISOFORM` extracts a canonical subset from the full `--gtf` (keeps the isoform with the longest CDS per gene).
- Routing: `ch_canonical_gtf` drives Ribo-TISH, Ribotricer, Plastid metagene/psite, and anota2seq/deltaTE; `ch_gtf` (full) drives STAR genome alignment, Salmon, RiboCode, and riboWaltz (transcriptome-BAM consumers must use the same annotation the BAM was built against).
- `--canonical_gtf` is a recommended backbone, not a hard requirement: nothing enforces exactly one transcript per gene, so an annotation that occasionally lists more than one (e.g. MANE Plus Clinical alongside MANE Select) is handled without error.
- `ch_fasta_gtf` meta carries `id: 'reference'` so published filenames are stable regardless of canonical / hybrid origin.

## Housekeeping (folded)

- Swap `modules/local/gtf_filter` for `nf-core/custom/gtffilter` (templated `gtffilter.py` is a strict superset of `bin/filter_gtf.py`; adds gzip in/out and an optional FASTA-based chromosome filter).
- Drop `modules/local/gtf2bed` and `bin/gtf2bed`; never invoked, BED output had no consumer.
- Drop `bin/filter_gtf.py` in favour of the in-module template.

## Stacked PR notes

Second in the stack splitting #174, now based on `dev` (#176 has merged).

Closes #161

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #174, opened 2026-05-18
Riboseq modernisation: Phases 1-3 (#160-#171 + PRICE)

pinin4fjords:
## Summary

Modernisation umbrella PR for the riboseq pipeline. Brings issues #160-#171 to a single aggregation branch ready for review, ahead of being split into per-issue PRs.

Default-path behaviour is preserved for everything except #160 (the documented breaking change to `--te_quantification_method`). All new ORF callers and the extended-ORF / cross-caller catalogue paths are opt-in.

### Quantification defaults (#160)
- **Breaking default change.** Flip `--te_quantification_method` default from `alignment` (STAR + Salmon) to `plastid_psite` (in-frame P-site counts). Salmon's coverage / fragment-length / multi-mapping assumptions are inappropriate for short Ribo-seq footprints; in-frame P-site counts are the scientifically correct quantity. Two methods produce different count values; re-running an existing cohort on the new default will not reproduce prior matrices. Backward-compatible behaviour available via `--te_quantification_method alignment`.

### Annotation backbone (#161)
- Add `--canonical_gtf` for an explicit one-transcript-per-gene canonical annotation (MANE Select / Ensembl canonical). Used by ORF calling, riboWaltz P-site calibration, plastid P-site quantification, and translational-efficiency analysis. Falls back to `AGAT_SP_KEEP_LONGEST_ISOFORM` extraction from `--gtf` when not supplied. The full `--gtf` continues to drive genome-guided alignment.

### Ribotricer demotion (#163)
- Replace `--skip_ribotricer` with `--run_ribotricer` (default `false`). Default caller set is now Ribo-TISH + RiboCode; agreement logic is parameterised on the runtime-enabled caller set. Ribotricer's score column drops out of cross-caller rank aggregation when not enabled.

### Novel transcript discovery (#164)
- StringTie reference-guided novel-transcript discovery. Prefer RNA-seq BAMs; fall back to Ribo-seq BAMs with tightened parameters (`--stringtie_ribo_fallback_args`) and a runtime warning.
- `--novel_gtf` to bypass StringTie with a user-supplied GTF.
- Class-code filter via `--stringtie_class_codes` (default `u`, intergenic only).
- Optional strand-aware rRNA/repeat blacklist via `--rrna_blacklist` (`bedtools intersect -v -s`).
- Hybrid GTF (`<outdir>/stringtie/hybrid_reference.gtf`) by concatenating canonical backbone + filtered novel transcripts. Exposed as the `hybrid_gtf` workflow channel.

[...]

PR #178, opened 2026-05-22
feat: demote Ribotricer to opt-in --run_ribotricer

### LINKED ISSUE #163: fix: demote Ribotricer to opt-in ORF caller
# fix: demote ribotricer to opt-in ORF caller

**Repo:** nf-core/riboseq
**Size:** S
**Blocked by:** —

## Background

The nf-core/riboseq benchmark (FK/NGB, May 2026; 6 biological replicates of 63F ± OSKM partial reprogramming, day 0 and day 8) evaluated six Ribo-seq ORF callers at genome-wide scale. Ribotricer (v1.5.0) scored poorly on replicability: mean Spearman correlation of ORF scores across replicate pairs = 0.288, despite mean Jaccard set overlap = 0.770. The diagnostic pattern — consistent *sets* but inconsistent *ranks* — means its score column is unreliable for any downstream ranking or multi-caller aggregation step.

The pipeline currently runs Ribotricer as one of three default ORF callers (Ribo-TISH, Ribotricer, RiboCode). Including a rank-unstable caller in the default multi-caller agreement step corrupts the consensus score signal.

Ribotricer is maintained and installable (Bioconda); the problem is algorithmic on this dataset. Users with specific reasons to run it (e.g., comparing against prior published results, or exploring its binary call set independent of scores) should retain the option.

## Proposed changes

1. Move Ribotricer behind an opt-in flag: `--run_ribotricer` (default `false`).
2. When `--run_ribotricer false` (default): remove it from the multi-caller agreement step entirely. The default caller set becomes Ribo-TISH + RiboCode until Rp-Bp is added (issue #14).
3. When `--run_ribotricer true`: run as before, but exclude its score column from the cross-caller rank aggregation. Its binary call set (ORF detected / not detected) can still contribute to the union catalogue used in issue #09.
4. Emit a `log.warn` when `--run_ribotricer true` is set, noting the benchmark finding and advising users not to use its scores for ranking.
5. Update multi-caller agreement logic (issue #07) to handle a variable caller set (2 callers by default, 3 if `--run_ribotricer` is set).
6. Add a CHANGELOG entry documenting the behaviour change (opt-out → opt-in).

## Notes

No module removal required — the existing `modules/nf-core/ribotricer/` import is retained; only the default invocation logic changes.

## References

- Benchmark report: FK_bioinf_projects/260511_Ribo_benchmarks (Krueger + García Bediaga, May 2026)
- Ribotricer: Supplementary benchmarking results retained for documentation

pinin4fjords:
## Summary

Demote Ribotricer from the default ORF-caller set to opt-in via `--run_ribotricer` (default `false`). The default caller set is now Ribo-TISH + RiboCode.

## Changes

- Replace `--skip_ribotricer` (default `false`) with `--run_ribotricer` (default `false`). Anyone who relied on the previous default now needs to set `--run_ribotricer true`.
- Runtime warning when Ribotricer is enabled, noting the rank-instability caveat.
- Cross-caller agreement logic parameterised on the runtime-enabled caller set so the threshold/majority math works whether 2, 3, or more callers are active. Ribotricer's score is excluded from rank aggregation; its binary calls are not.
- Agreement-majority calc uses `intdiv(2) + 1` instead of `Math.floor(n/2.0d)` so Nextflow 26.04's strict config parser accepts it.
- Document the opt-in status and the precision-weighted default agreement in `docs/usage.md`.

## Design note

The benchmark (FK/NGB, May 2026; 6 biological replicates) found two distinct things about Ribotricer: its ORF-score column is rank-unstable across replicates (mean Spearman 0.288), while its binary call set is reasonably reproducible (mean Jaccard 0.770). The opt-in decision weighs both: the unreliable signal (the score) is excluded from rank aggregation, and Ribotricer is dropped from the default not because its calls are unreproducible but because it is a third periodicity/frame-based caller (like Ribo-TISH and RiboCode) rather than an independent line of evidence, so it adds little methodological diversity for the runtime cost. Its binary calls still contribute when explicitly enabled.

Consequently the default two-caller agreement (Ribo-TISH + RiboCode, both must agree) is precision-weighted: conservative, and may omit ORFs only one caller detects. Enable `--run_ribotricer` for a third vote and broader recall.

## Stacked PR notes

Part of the stack splitting #174. Parent #177 (`feat/161-canonical-gtf`) has merged, so this now targets `dev` directly and includes a merge of `dev` to pick up the plastid-default and canonical-GTF work.

Closes #163


FelixKrueger:
After our benchmark of ORF callers, this move seems warranted.

FelixKrueger (approved):

Issue #171, opened 2026-05-14
feat: second STAR pass against hybrid transcriptome to give RiboCode equal novel-locus coverage


pinin4fjords:
## Summary

RiboCode requires a transcriptome-coordinate BAM (STAR `--quantMode TranscriptomeSAM`) built against the reference transcriptome at alignment time. Novel StringTie transcripts added post-hoc are invisible. This issue restores three-caller parity for novel ORFs by rebuilding the transcriptome index to include novel sequences and re-aligning.

**Blocked by:** #165 (hybrid GTF wired and validated). This is a Phase 3 issue — undertake only after Phase 2 (#165) is validated in production.

## Rationale

Under the current plan, canonical ORFs get three callers (Ribo-TISH, Ribotricer→Rp-Bp, RiboCode) while novel ORFs get only two (Ribo-TISH + Rp-Bp when active). The pipeline's confidence model rests on multi-caller agreement; that asymmetry is scientifically unjustified for precisely the ORFs we are most uncertain about.

## Implementation

```bash
# 1. Extract novel transcript sequences from genome
gffread novel_intergenic.gtf -g genome.fa -w novel_transcripts.fa

# 2. Build hybrid transcriptome FASTA
cat reference_transcriptome.fa novel_transcripts.fa > hybrid_transcriptome.fa

# 3. Build new STAR index against hybrid transcriptome
STAR --runMode genomeGenerate \
     --genomeFastaFiles genome.fa \
     --sjdbGTFfile hybrid_reference.gtf \
     --genomeDir hybrid_star_index/

# 4. Re-align with TranscriptomeSAM output
STAR --quantMode TranscriptomeSAM \
     --genomeDir hybrid_star_index/ \
     ...
```

The hybrid transcriptome BAM is used for RiboCode and riboWaltz only. Ribo-TISH and Plastid continue using the genome-coordinate BAM.

## Cost

~2× alignment compute cost per sample. Use `--extended_orf_analysis` gating so it is only triggered when needed. Cache the hybrid STAR index across samples within a run.

## References

- RiboCode: [Xiao et al. NAR 2018](https://doi.org/10.1093/nar/gky179)
- gffread: [Pertea & Pertea, F1000Research 2020](https://doi.org/10.12688/f1000research.23297.2)

pinin4fjords:
Resolved by #184, merged into `dev`.

Issue #165, opened 2026-05-14
feat: wire hybrid GTF (canonical + novel intergenic) into ORF callers


pinin4fjords:
## Summary

PR #158 produces a merged StringTie GTF but does not feed it into the ORF callers — novel transcript discovery currently has no effect on biological outputs. This issue wires the hybrid GTF (canonical backbone from #161 + class `u` novel transcripts from #164) into Ribo-TISH and Ribotricer.

**Blocked by:** #161 (canonical backbone), #164 (StringTie/gffcompare filter), nf-core/modules#11644 (ribotish/predict `-a` flag), #162 (ribotish/quality investigation). Gated on `--extended_orf_analysis`.

## Architecture decision: transcriptome-BAM constraint

STAR `--quantMode TranscriptomeSAM` produces a BAM keyed to the reference transcriptome at alignment time. **RiboCode** and **riboWaltz** consume this transcriptome BAM — novel StringTie sequences not present at alignment time are invisible to them.

**Resolution:**
- **riboWaltz:** canonical annotation only, permanently. It is a QC tool; hybrid GTF input has no scientific value and would degrade CDS-diagnostic plot readability.
- **RiboCode:** canonical only in this phase. Equal novel-locus coverage for RiboCode requires a second STAR alignment against a hybrid transcriptome — addressed separately in issue #168 (second STAR pass). That issue must be filed and addressed to restore three-caller parity for novel ORFs.
- **Ribo-TISH + Ribotricer:** genome-BAM tools — can accept any GTF directly. Updated in this issue.

## Hybrid GTF construction

```
canonical_backbone.gtf   (MANE / Ensembl_canonical, issue #161)
  + novel_intergenic.gtf (StringTie class 'u', issue #164)
  = hybrid_reference.gtf
```

Emit `hybrid_reference.gtf` as a published output at `<outdir>/stringtie/hybrid_reference.gtf`.

## Per-tool wiring

**Ribo-TISH `predict`** (requires nf-core/modules#11644):
```
-g novel_intergenic.gtf        # discovery target
-a canonical_backbone.gtf      # background model + ORF classification
```
Or pass `hybrid_reference.gtf` to `-g` (simpler; slightly less clean separation).

**Ribotricer `prepare-orfs`:** pass `hybrid_reference.gtf` directly. No module change needed — Ribotricer classifies all CDS-absent transcripts as `novel` automatically.

[...]

Issue #161, opened 2026-05-14
feat: support one-transcript-per-gene canonical annotation backbone (MANE Select / Ensembl canonical)


pinin4fjords:
## Summary

The pipeline currently uses the full multi-isoform Ensembl/GENCODE GTF. For Ribo-seq this creates unavoidable multi-mapping: only ~7% of reads map uniquely; 64% multi-map due to shared exons across isoforms (Ribomap 2016). Restricting to one representative transcript per gene eliminates shared-exon ambiguity by design.

## Proposed change

Add an optional `--canonical_gtf` parameter. When supplied, this replaces the full reference GTF for annotation purposes (ORF calling, P-site calibration, DTE). The full GTF is still used for genome-guided alignment.

### Recommended sources by organism

| Organism | Source | Extraction |
|---|---|---|
| Human (GRCh38) | [MANE Select GTF](https://ftp.ncbi.nlm.nih.gov/refseq/MANE/MANE_human/current/) | Use directly |
| Any Ensembl organism (release ≥104) | Ensembl GTF | `grep 'tag "Ensembl_canonical"' input.gtf > canonical.gtf` |
| Any organism (fallback) | Full GTF + AGAT | `agat_sp_keep_longest_isoform.pl` (nf-core module exists) |

When `--canonical_gtf` is not supplied, the pipeline falls back to existing full-GTF behaviour.

## Implementation notes

- MANE Select GTF includes `gene`, `transcript`/`mRNA`, `exon`, `CDS` features — compatible with all downstream tools without modification.
- `Ensembl_canonical` approach requires a single `grep`; no new dependencies.
- AGAT fallback requires adding `agat/spkeeplongestisoform` as a conditional module import.
- **Scope:** MANE Select is the backbone for gene-level P-site counting and DTE (issues #160, #10 Tier-1). The full multi-isoform annotation is retained separately as input for the extended ORF discovery workflow (issues #7, #8, #9) where non-MANE isoforms matter for uORF and alternative-ORF discovery. These are two separate annotation inputs.

## References

- Morales et al. (2022) Nature 604:310. [DOI: 10.1038/s41586-022-04558-8](https://doi.org/10.1038/s41586-022-04558-8)
- Ribomap: [Bioinformatics 2016, 32(12):1880](https://academic.oup.com/bioinformatics/article/32/12/1880/1744291)
- [Ensembl canonical docs](https://www.ensembl.org/info/genome/genebuild/canonical.html)
[...]

PR #182, opened 2026-05-22
feat: StringTie novel transcript discovery + hybrid GTF

### LINKED ISSUE #164: feat: StringTie novel transcript discovery — RNA-seq BAMs preferred, user-supplied GTF bypass, intergenic-only filter
## Summary

Three refinements to PR #158 (StringTie novel transcript discovery):

1. Accept user-supplied `--novel_gtf` to bypass assembly entirely
2. Use RNA-seq BAMs for StringTie when available (Ribo-seq BAMs are a fallback)
3. Filter output to gffcompare class `u` (intergenic-only) before passing downstream

**Blocked by:** PR #158 merged. Part of `--extended_orf_analysis` opt-in flag.

## 1. User-supplied GTF (`--novel_gtf`)

Users with a pre-built cell-type-specific annotation (e.g. from long-read RNA-seq or a published tissue-specific catalogue) can supply it directly via `--novel_gtf`, skipping StringTie assembly. The supplied GTF is passed through the same gffcompare class `u` filter and used identically downstream. StringTie only runs when `--novel_gtf` is absent.

## 2. Prefer RNA-seq BAMs for StringTie

Every published microprotein workflow assembles from RNA-seq, not Ribo-seq: RIBOSS ([Lim et al. 2025](https://doi.org/10.1093/bib/bbaf164)), Rp3 ([Vieira de Souza et al. 2024](https://doi.org/10.1038/s41467-024-50301-4)), Wu et al. tomato ([2019](https://doi.org/10.1104/pp.19.00541)), ggRibo Arabidopsis protocol (PMID: 41241901). Ribo-seq footprints produce shallow, sparse overlap graphs in StringTie with non-uniform P-site-biased coverage, leading to fragmented assemblies and false-positive single-exon "transcripts" from ribosome pause sites.

```nextflow
// in workflows/riboseq/main.nf
ch_stringtie_input_bam = rnaseq_bams_available
    ? ch_rnaseq_genome_bam
    : ch_genome_bam  // Ribo-seq fallback
```

When using the Ribo-seq fallback, emit a pipeline warning and apply tightened `ext.args`:

| Parameter | RNA-seq default | Ribo-seq fallback |
|---|---|---|
| `-m` (min transcript length) | 200 bp | 100 bp |
| `-c` (min coverage) | 1.0 | 5.0 reads/bp |
| `-j` (min junction reads) | 1.0 | 3.0 |
| `-f` (min isoform fraction) | 0.01 | 0.05 |
| `-g` (max gap) | 50 bp | 100 bp |

Post-assembly: discard transcripts with TPM < 0.5 (following Wu et al. 2019).

These Ribo-seq fallback parameters are first-pass empirical defaults — not literature-validated for Ribo-seq BAM input. Treat as a starting point pending quality assessment on real data (assembly size, length distribution, class code breakdown).

## 3. gffcompare + class code filter

After `STRINGTIE_MERGE`:

```bash
# 1. Run gffcompare
gffcompare -r reference.gtf stringtie_merged.gtf -o gffcmp

# 2. Filter to class 'u' (zero overlap with any annotated gene on either strand)
awk '$3=="transcript"||$3=="exon"' gffcmp.annotated.gtf \
    | grep 'class_code "u"' > novel_intergenic.gtf

# 3. Concatenate with canonical backbone (issue #161)
cat canonical_backbone.gtf novel_intergenic.gtf > hybrid_reference.gtf
```

Add `--stringtie_class_codes` parameter (default `"u"`). For fully stranded RNA-seq + stranded Ribo-seq, class `x` (antisense) can be added via `--stringtie_class_codes "u,x"` to recover genuinely translated antisense transcripts.

Post-filter: optionally intersect with `--rrna_blacklist` BED to remove rRNA/repeat artefacts. Make blacklist step conditional (skip silently if absent with a warning) — most non-human organisms lack a published equivalent.

## References

- GffCompare: [Pertea et al. F1000Research 2020](https://doi.org/10.12688/f1000research.23297.2)
- [gffcompare class codes](https://ccb.jhu.edu/software/stringtie/gffcompare.shtml)
- [StringTie manual](https://ccb.jhu.edu/software/stringtie/index.shtml?t=manual)

pinin4fjords:
## Summary

Adds reference-guided novel-transcript assembly via StringTie (preferring RNA-seq BAMs, falling back to Ribo-seq with tightened defaults) or a user-supplied annotation via `--novel_gtf`. Either source flows through a common filtering chain (gffcompare class-code filter, optional strand-aware rRNA/repeat blacklist, concat with the canonical backbone) and produces a hybrid annotation at `<outdir>/stringtie/hybrid_reference.gtf`.

This is the foundation for #165 / #171 / extended ORF discovery. With the default `--skip_stringtie true` and no `--novel_gtf`, the new `hybrid_gtf` workflow channel equals `ch_canonical_gtf` from #161, so downstream wiring is uniform whether novel discovery is on or off.

## Changes

- New local subworkflow `NOVEL_TRANSCRIPT_DISCOVERY` chaining `bam_stringtie_merge` (per-sample StringTie + merge) and the upstream `gtf_hybridmerge_gffcompare` subworkflow (gffcompare + class-code filter + optional blacklist intersect + concat with canonical).
- Install upstream modules: `stringtie/stringtie`, `stringtie/merge`, `gffcompare`, `bedtools/intersect`.
- Install upstream subworkflows: `bam_stringtie_merge`, `gtf_hybridmerge_gffcompare` (PR nf-core/modules#11729, merged).
- New params: `--skip_stringtie` (default `true`), `--novel_gtf`, `--stringtie_class_codes` (default `'u'`, intergenic only), `--rrna_blacklist`, `--extra_stringtie_args`, `--extra_stringtie_merge_args`, `--stringtie_ribo_fallback_args` (default `'-m 100 -c 5 -j 3 -f 0.05 -g 100'` to tighten the Ribo-seq fallback).
- Workflow emits `hybrid_gtf` from the riboseq workflow.

## Scientific basis

Each design choice was checked against the ribosome-profiling / microprotein-discovery literature (primary sources linked).

- **StringTie reference-guided assembly for discovery.** Running StringTie with `-G` and without `-e` is the standard discovery mode and the engine used across the microprotein field: RIBOSS ([Lim et al., Brief Bioinform 2025, 26(2):bbaf164](https://academic.oup.com/bib/article/26/2/bbaf164/8113023)), Rp3 ([Vieira de Souza et al., Nat Commun 2024, 15:6839](https://www.nature.com/articles/s41467-024-50301-4)), and the tomato translatome study ([Wu et al., Plant Physiol 2019; PMID 31248964](https://pubmed.ncbi.nlm.nih.gov/31248964/)).
- **RNA-seq BAMs preferred; Ribo-seq only as a warned fallback.** Every published workflow assembles transcript structure from RNA-seq or long-read RNA-seq, never from Ribo-seq footprints, which are CDS-concentrated and non-uniform (Wu et al. report 85.5% of footprints in-frame on CDS), leaving UTR / non-coding structure poorly resolved. RIBOSS states it "assembles transcriptomes from long and short reads using StringTie", with Ribo-seq used only downstream ([PMC11994033](https://pmc.ncbi.nlm.nih.gov/articles/PMC11994033/)). There is no published validation of Ribo-seq-only assembly or of the tightened fallback parameters, hence the runtime warning and the non-default status.
- **`--novel_gtf` bypass.** Long-read-derived or curated annotations are the strongest input for transcript structure; RIBOSS likewise accepts a pre-assembled GTF directly.
- **gffcompare class code `u` by default.** `u` is the only code with zero overlap to any annotated gene on either strand, i.e. unambiguously novel intergenic ([Pertea & Pertea, F1000Research 2020; PMC7222033](https://pmc.ncbi.nlm.nih.gov/articles/PMC7222033/)). Restricting to `u` favours precision; `x` (antisense exonic overlap) is offered opt-in for stranded libraries only.
- **Strand-aware rRNA/repeat blacklist (`bedtools intersect -v -s`).** rRNA loci are usually absent from coding-gene annotations and so surface as false class-`u` novel transcripts; a strand-aware subtractive intersect removes them, which matters most on the Ribo-seq fallback path.
- **Hybrid GTF feeding genome-coordinate ORF callers.** The hybrid annotation is consumed by Ribo-TISH `predict` ([Zhang et al., Nat Commun 2017, 8:1749](https://www.nature.com/articles/s41467-017-01981-8)) and Ribotricer ([Choudhary et al., Bioinformatics 2020, 36(7):2053](https://academic.oup.com/bioinformatics/article/36/7/2053/5637228)), both of which accept mixed CDS-bearing / CDS-absent GTFs. That wiring lands in #165.

Discovery here is structural only; whether a novel locus is genuinely translated is decided downstream by the periodicity-based ORF callers. Deferred precision refinements (an abundance/TPM floor per Wu et al., single-exon scrutiny, and benchmarking the Ribo-seq fallback) are tracked in #191.

## Caveats

- Ribo-seq StringTie defaults are first-pass empirical, not literature-validated. The pipeline warns when the fallback is active.
- Default `--stringtie_class_codes = u` keeps only unambiguously intergenic transcripts. Stranded users can opt into `'u,x'` to recover antisense; unsafe for non-stranded protocols.
- `--rrna_blacklist` is silently skipped when not supplied.

## Stacked PR notes

Part of the stack splitting #174. The parent (#181) has merged, so this now targets `dev`; the integration test (`tests/novel_gtf.nf.test`) exercises the `--novel_gtf` bypass and asserts the hybrid GTF is produced.

Closes #164


nextflow.config
```diff
@@ -131,6 +131,15 @@ params {
     // TE quantification method
     te_quantification_method          = 'plastid_psite'
 
+    // StringTie novel transcript discovery
```
iraiosub, line 134: I wonder given ribo-seq data isn't ideal for transcript assembly, whether instead of using it as a fallback when rna-seq is missing, we could instead have an explicit  param like `--allow_riboseq_stringtie` for it. So users should explicitly enable/opt-in this fallback while being aware of the caveats.

bediagan: I echo Iraia's point. Since Ribo-seq data isn't built for transcript assembly, it shouldn't silently replace missing RNA-seq data. An explicit parameter like --allow_riboseq_stringtie would at least force an intentional opt-in.

However, to keep the pipeline robust and aligned with best practices, should we just skip adding it entirely? If it's a bad use case, not supporting it at all might be the cleanest approach.

CHANGELOG.md
```diff
@@ -68,6 +69,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 |                     | `--skip_coverage_tracks`                 |
 |                     | `--ribodetector_chunk_size`              |
 |                     | `--canonical_gtf`                        |
+|                     | `--skip_stringtie`                       |
+|                     | `--novel_gtf`                            |
+|                     | `--stringtie_class_codes`                |
```
suhrig: Wouldn't `gffcompare_class_codes` be a more appropriate name?

subworkflows/local/novel_transcript_discovery/main.nf
```diff
@@ -0,0 +1,102 @@
...
+    ribo_fallback_args  // val: string                          - StringTie args for Ribo-seq fallback (for warning)
+
+    main:
+
+    ch_novel_pre_filter = Channel.empty()
+
+    if (user_novel_gtf) {
+        ch_novel_pre_filter = Channel
+            .fromPath(user_novel_gtf, checkIfExists: true)
+            .map { gtf -> [ [id: 'stringtie_merge'], gtf ] }
```
suhrig: Should we name the ID differently? StringTie was not involved in the creation of this GTF after all.

conf/modules.config
```diff
@@ -1076,4 +1076,77 @@ process {
...
+            path: { "${params.outdir}/stringtie" },
+            mode: params.publish_dir_mode,
+            saveAs: { filename -> filename.equals('versions.yml') ? null : filename }
+        ]
+    }
+
+    // Subworkflow's bundled nextflow.config defaults BEDTOOLS_INTERSECT to '-v';
+    // append '-s' to match the pre-swap strand-aware blacklist subtraction.
+    withName: '.*:GTF_HYBRIDMERGE_GFFCOMPARE:BEDTOOLS_INTERSECT' {
+        ext.args   = '-v -s'
```
suhrig: Is it technically possible to set `-s` conditionally, i.e., only when the library is actually stranded? Or can't this be done, because at the time of `ext.args` definition it is unclear whether RNA or Ribo will serve as a basis for transcript discovery and the strandedness cannot be determined reliably, hence?

nextflow.config
suhrig: I'm leaning towards removal of this feature, too. I doubt that transcript discovery from 30bp read length is good practice, to the extent where it may do more harm than good.

pinin4fjords: OK, agreed, will do

CHANGELOG.md
pinin4fjords: Done - renamed to `--gffcompare_class_codes` throughout (schema, config, subworkflow, docs, CHANGELOG).

subworkflows/local/novel_transcript_discovery/main.nf
pinin4fjords: Good catch - changed to `id: 'novel_gtf'` on the user-supplied path. The StringTie merge path keeps `id: 'stringtie_merge'`.

conf/modules.config
pinin4fjords: Yes, via an `ext.args` closure - and it turned out to be a correctness bug rather than just a nice-to-have: hardcoding `-s` was wrong for unstranded libraries (strand-match required, so `bedtools intersect` would silently miss blacklist hits on the opposite strand). Fixed by threading strandedness from the BAM meta into the GTF meta via `.combine(ch_strandedness)` in the subworkflow, then using a closure: `{ meta.strandedness != 'unstranded' ? '-v -s' : '-v' }`.

suhrig (approved):

iraiosub (approved):

Issue #168, opened 2026-05-14
feat: ORF-level differential translation analysis (DT, DTE, and DOU)


pinin4fjords:
## Summary

Extend the DTE pipeline from gene level to ORF level. The current tools (anota2seq, deltaTE) operate at gene level; this issue adds ORF-level DT/DTE and lays the groundwork for DOTSeq when it reaches Bioconductor stable.

**Blocked by:** #166 (ORF-level P-site counts), #167 (merged ORF catalogue).

## Terminology

- **Differential Translation (DT):** change in absolute ribosome occupancy. DESeq2 on Ribo-seq counts directly. Does not require RNA-seq.
- **Differential Translation Efficiency (DTE):** change in translation normalised against mRNA abundance. Requires paired Ribo-seq + RNA-seq. Tools: anota2seq, deltaTE, DOTSeq DTE.
- **Differential ORF Usage (DOU):** relative contribution of each ORF to gene's total translation output; beta-binomial GLM (DOTSeq). Does not require an absolute RNA-seq denominator. Detects uORF/mORF balance shifts invisible to gene-level DTE.

## Tier 1 — Gene-level DTE (backward compatible)

Keep existing anota2seq and deltaTE modules. Pre-aggregate ORF P-site counts to gene level **using main CDS ORFs only** before these tools:

```r
gene_level_counts <- orf_psite_counts %>%
    left_join(orf_to_gene_mapping, by = "orf_id") %>%
    filter(orf_type == "CDS") %>%   # main CDS ORFs only — NOT uORFs or dORFs
    group_by(gene_id, sample) %>%
    summarise(counts = sum(counts), .groups = "drop") %>%
    pivot_wider(names_from = sample, values_from = counts)
```

Summing uORF counts into the gene total contaminates the TE numerator. Use `orf_to_gene.tsv` from issue #167.

**anota2seq distribution note:** P-site counts are sparser and more zero-inflated than Salmon TPM-derived counts. Default anota2seq filtering thresholds (minimum counts per gene) were calibrated on RNA-seq-like distributions. Validate empirically and document adjusted thresholds for P-site input before releasing.

## Tier 2 — ORF-level DTE via DESeq2 interaction model

Apply `~ condition + seq_type + condition:seq_type` per ORF. The `condition:seq_type` term gives per-ORF TE change. Extrapolation from gene-level published precedent (Chothani et al. 2019); P-site counts at ORF level are often sparse.

[...]

Issue #167, opened 2026-05-14
feat: cross-sample ORF catalogue — merge, deduplicate, and re-quantify across all samples


pinin4fjords:
## Summary

ORF callers run per-sample producing per-sample ORF lists. For a consistent count matrix, these must be merged into a single run-level catalogue before quantification — analogous to merging ATAC-seq peaks into a consensus set before re-quantification.

**Blocked by:** #166 (ORF-level quantification, which defines the ORF catalogue format). Part of `--extended_orf_analysis`.

### Canonical ORFs (>100 aa)

**Two-path approach** to handle spliced ORFs:
- **Annotated multi-exon CDS ORFs:** collapse by shared transcript ID across samples. These are already uniquely identified; coordinate merging would incorrectly merge ORFs sharing genomic span but with different intron structures.
- **Single-exon novel intergenic ORFs (class `u`):** coordinate-based strand-aware merge:
  ```bash
  cat sample*.bed | sort -k1,1 -k2,2n | bedtools merge -s > merged_canonical.bed
  ```

Near-identical ORFs with caller-specific TIS uncertainty: collapse using **99% amino acid identity + shared stop codon** — the criterion used by the GENCODE Ribo-seq ORF catalogue (Ruiz-Orera et al. Nat Biotechnol 2022). Not a nucleotide-window heuristic.

### Small ORFs (≤100 aa — the smORF threshold)

The 100 aa / 300 nt cutoff is the field standard: sORFs.org (NAR 2016), GENCODE Ribo-seq catalogue, SmProt v2. After coordinate merge, apply sequence-based deduplication — smORFs in repetitive regions may be called at multiple genomic copies representing the same peptide:

```bash
# CD-HIT at 100% identity
cd-hit -i merged_smorfs.faa -o deduped.faa -c 1.00 -n 5 -l 10 -M 4000 -T 8

# Or MMseqs2 (faster for large catalogues; --cluster-mode 2 for short peptides)
mmseqs easy-cluster merged_smorfs.faa result tmp/ --min-seq-id 1.0 --cov-mode 0 -c 1.0 --cluster-mode 2
```

### Recurrence filter (optional)

Require each ORF called in ≥2 samples (or ≥1 per condition) before catalogue inclusion. Reduces single-sample false positives without hard intersection.

### Re-quantification

All samples re-quantified against the merged catalogue (issue #166). Samples where an ORF was not called have zero counts — zero is a valid observation, not missing data.

[...]

PR #188, opened 2026-05-22
feat: per-ORF P-site quantification

### LINKED ISSUE #166: feat: ORF-level P-site quantification — replace gene-level counting with per-ORF counts
## Summary

Replace gene-level P-site aggregation (`QUANTIFY_INFRAME_PSITE_PLASTID`) with per-ORF counting. Each ORF (main CDS, uORF, novel smORF) gets its own count row, enabling ORF-level DTE and uORF-mediated regulation detection.

**Blocked by:** #161 (canonical backbone), #165 (hybrid GTF wired into callers). Part of `--extended_orf_analysis`.

## Current behaviour

`bin/gtf_to_inframe_psites.awk` is called with `FEATURE=gene` (see `workflows/riboseq/main.nf` ~line 547), which aggregates all in-frame P-sites to the gene level. The per-ORF version requires `FEATURE=transcript` (or ORF-ID-based handling) so each ORF gets its own coordinate set.

## Implementation options

**Option A — Extend existing Plastid wiggle approach (recommended):** Keep `PLASTID_MAKE_WIGGLE`. Replace `gtf_to_inframe_psites.awk` with a script that accepts the merged ORF catalogue (BED12 or GTF) and emits codon-start positions using the ORF's own ATG as frame 0 (independent of GTF `phase` field). Reuse the `bedtools intersect + groupby` pattern. **Only option giving true codon-start counting.**

**Option B — Plastid `counts_in_region`:** `counts_in_region --fiveprime_variable --offset p_offsets.txt --annotation_files merged_orfs.bed`. Source confirmed: calls `numpy.nansum(ivc.get_masked_counts(ga))` — sums **all** P-site positions in the span, not codon-start only. Less precise than A but avoids custom post-processing.

**Option C — DOTSeq flattening + featureCounts:** `orf_to_gtf.py` creates a non-overlapping flattened GTF; featureCounts counts full RPF reads per ORF. Handles same-frame overlapping ORFs correctly; requires DOTSeq (currently Bioconductor dev only).

## ORF ID scheme

Stable identifiers encoding enough information for reproducibility:
- Canonical CDS: `ENST00000123456.5_CDS_100_500`
- Non-canonical: `chr1_+_10000_10150_frame0`

Row names in the count matrix must be consistent between Ribo-seq (numerator) and RNA-seq (denominator) matrices for DTE.

## RNA-seq denominator assignment (for DTE)

| ORF class | Denominator |
|---|---|
| Canonical CDS ORF | Gene-level RNA-seq count |
| uORF / dORF on annotated transcript | Transcript-level Salmon count |
| ORF on StringTie novel transcript | Transcript-level count (Salmon against hybrid GTF) |
| Novel intergenic ORF, no host transcript | None — counts only, excluded from DTE |

## Same-frame overlap caveat

Two ORFs in the **same frame** that overlap (e.g. N-terminally extended CDS variants) will both claim the same in-frame P-site under Options A and B. This is rare in canonical annotations but real for some ORF classes. Add explicit test cases for same-frame overlap to document behaviour. Option C (DOTSeq flattening) handles this automatically.

## anota2seq compatibility note

P-site counts are sparser and more zero-inflated than Salmon TPM-derived counts. anota2seq's APV regression default filtering thresholds were calibrated on RNA-seq-like distributions — validate empirically and adjust minimum-count thresholds for P-site input before releasing (tracked in issue #167).

## Cross-tool consensus: fuzzy overlap required

Benchmark finding: strict (chrom, start, end, strand) coordinate-equality cross-tool Jaccard = 0.000–0.057 due to alt-start ambiguity. ORF catalogue merging (issue #167) must use fuzzy overlap (≥80% reciprocal bedtools-style), not coordinate equality.

## Outputs

- `orf_psite_counts.tsv`: ORF × sample raw count matrix
- `orf_catalogue.gtf` / `.bed12`: merged ORF catalogue used for counting
- `orf_to_gene.tsv`: ORF ID → host gene ID mapping (for gene-level pre-aggregation, issue #167)

## References

- ORFquant (23% dominant ORF on non-principal isoform): [Calviello et al. Nat Struct Mol Biol 2020](https://doi.org/10.1038/s41594-020-0450-4)
- DOTSeq: [Lim & Chieng, bioRxiv 2025](https://doi.org/10.1101/2025.09.24.678201)
- Plastid `counts_in_region`: [gene expression tutorial](https://plastid.readthedocs.io/en/latest/examples/gene_expression.html)
- Current code: `modules/local/quantify_inframe_psite_plastid/`, `bin/gtf_to_inframe_psites.awk`

pinin4fjords:
## Summary

Adds a per-ORF in-frame P-site quantification path that runs additively to the existing gene-level plastid path. Expands the cohort ORF catalogue (#167) BED12 into codon-start BED6 positions, runs `bedtools intersect` per sample against the plastid wiggle tracks, and assembles a zero-filled ORF x sample count matrix. The matrix is the input for the per-ORF DTE step tracked in #168; gene-level quantification and DTE are unchanged.

## Changes

- New local subworkflow `QUANTIFY_ORF_PSITE` chaining `CUSTOM_BED12CODONPOSITIONS` (cohort-level codon-start expansion) + `QUANTIFY_INFRAME_PSITE_ORF` (per-sample strand-aware bedtools intersect + groupby) + `ORF_COUNT_MATRIX` (pivot to ORF x sample matrix, zero-filled for ORFs absent from a sample).
- New local modules `quantify_inframe_psite_orf` and `orf_count_matrix`; new nf-core module `custom/bed12codonpositions`.
- Workflow integration in the plastid block: `QUANTIFY_ORF_PSITE` runs when `--extended_orf_analysis true`, at least one ORF caller is enabled, and plastid is not skipped. A runtime warning is emitted when `--extended_orf_analysis true` is combined with `--skip_plastid true` (the catalogue is still built; the per-ORF matrix is skipped).
- `conf/modules.config`: outputs published under `<outdir>/orf_quantification/` (`*.orf_inframe_psites.bed`, `orf_psite_counts.tsv`).
- `docs/usage.md`, `docs/output.md`, `CHANGELOG.md` updated.

No new pipeline parameters: the feature is gated entirely on the existing `--extended_orf_analysis`, `--skip_plastid` and ORF-caller flags.

## Codon-start counting (not span counting)

Only P-sites falling exactly on the first position of each in-frame codon are counted (positions 0, 3, 6, ... relative to each ORF's own start codon). This maximises in-frame purity at the cost of ~1/3 the raw counts vs span counting. The frame is defined by each ORF's own ATG (A of ATG = frame 0), not by the GTF `phase` field, so novel transcripts and non-canonical starts are handled correctly.

## Validation

In-frame correctness was verified empirically against the genome and the P-site tracks, not just via snapshot: an independent re-implementation of the codon walk + intersect reproduces `orf_psite_counts.tsv` exactly, and frame-0 P-sites are enriched 2.07:1 over the mean of frames 1/2 with 90% of catalogue ORFs individually frame-0-dominant, confirming the counts capture genuine triplet-periodic translation (chr20 test depth; enrichment is stronger at full scale). Correct framing depends on the ribotish `GenomePos` coordinate fix in nf-core/modules#12173 (re-pinned to `dev` via #193); before that fix the periodicity fell one base over in frame 1.

## Stacked PR notes

Part of the stack splitting #174, targeting `dev`. The per-ORF DTE consumer (`orf_count_matrix` workflow emit, DOTSeq / anota2seq / deltaTE at ORF level) lands in #168.

Closes #166


modules/local/quantify_inframe_psite_orf/main.nf
```diff
@@ -0,0 +1,56 @@
...
+        sed 's/\$/\\t0\\t-/' "$reverse_bedgraph"
+    ) |
+
+    # intersect codon-start positions with stranded P-site density,
+    # then aggregate counts per ORF id (column 4 of the inframe psites BED)
+    bedtools intersect -wa -wb -a $inframe_psites -b /dev/stdin -s |
+    sort -k4,4 |
+    bedtools groupby -g 4 -c 10 -o sum |
+    sed 's/^/${meta.id}\\t/' \\
+    > "${meta.id}.orf_inframe_psite_counts.tsv"
```
suhrig: This is redundant to the gene-level in-frame quantification (as I am sure you are aware). UNLESS OVERLY COMPLICATED, it would be ideal if we could consolidate this, i.e., drop the gene-level quantification in favor of this implementation or the other way round. Just a suggestion ...

suhrig (approved):

modules/local/quantify_inframe_psite_orf/main.nf
pinin4fjords: You're quite right, sorry, must have got duplicated as part of the bigger rewrite without my noticing. Fixed now, using using module.

PR #189, opened 2026-05-22
feat: ORF-level differential translation

### LINKED ISSUE #168: feat: ORF-level differential translation analysis (DT, DTE, and DOU)

pinin4fjords:
## What this adds

Ribo-seq measures translation (ribosomes on mRNA); the pipeline already reports **translational efficiency (TE)** - ribosome occupancy relative to mRNA level - and how it changes between conditions, **per gene**. This PR adds the same analysis at **ORF resolution**, because a gene can regulate its protein output through small ORFs (e.g. upstream ORFs) that a whole-gene number hides. The gene-level DTE path is unchanged.

## What you get

When `--extended_orf_analysis`, an ORF caller, plastid counting and `--contrasts` are all active, the pipeline runs differential translation at ORF resolution using the same `--translational_efficiency_method` selector as the gene-level path:

- **anota2seq** (default) / **deltate** - the existing engines, now also run per ORF.
- **dotseq** - DOTSeq (ORF-level only), which additionally answers a second question that DTE alone can't.

| Metric | Question it answers |
| --- | --- |
| Per-ORF DTE | Did this ORF's translational efficiency change between conditions? |
| DOU (DOTSeq) | Did this ORF change its *share* of its gene's ribosomes? |

DTE catches absolute shifts; DOU catches redistribution *within* a gene - a uORF switching on to repress its main ORF - which a whole-gene view, or even per-ORF DTE, would miss.

## How it works

`DTE_COUNTS_PREP` builds one ORF-keyed count matrix: each ORF's Ribo-seq P-site counts joined to its host gene's Salmon RNA-seq counts. The gene's RNA is replicated across all its ORFs, because an individual ORF has no separately measurable mRNA of its own. ORFs with no host gene (novel intergenic) are dropped at the join. The selected method is then fit treating each ORF as a feature; `dotseq` additionally fits a per-gene beta-binomial model for DOU. `FILTER_COUNTS_CANONICAL` keeps the gene-level Ribo numerator and the RNA denominator on the same canonical gene IDs. Selecting `dotseq` without its prerequisites fails fast (`dotseqPrerequisitesError`).

## Limitations

Per-ORF results carry statistical caveats (documented in `docs/usage.md` and `docs/output.md`). Treat per-ORF p-values as a ranking to prioritise candidates, not strict significance:

- **Row independence.** Sibling ORFs from one gene share the same borrowed gene-level RNA denominator, so their rows are perfectly correlated, but anota2seq and deltaTE treat each ORF as independent. This is a biological identifiability limit: ORFs on the same mRNA have no separate RNA level.
- **Low-count ORFs.** uORFs / smORFs / low-abundance novel ORFs have sparse counts where DESeq2 dispersion and anota2seq's RVM are unreliable; check the diagnostic plots and tune via `--extra_orf_dte_args` / `--extra_anota2seq_run_args` / `--extra_dotseq_args`.
- **Multiple-testing scale.** Per-ORF runs have far more rows than per-gene runs (5-50x), so BH-adjusted p-values are more conservative by construction.
- **Overlapping-ORF quantification.** Same-frame overlapping ORFs double-count shared codon positions in the upstream P-site counts, which can distort DOU shares. The principled fix (an inferential-replicate method such as Fishpond/Swish on a bootstrapped ORF quantification) is out of scope, as it conflicts with the pipeline's deterministic in-frame P-site counting.

## Changes

- New modules: `dotseq/dotseq` (nf-core/modules master), local `dte_counts_prep`, and `assets/filter_counts_canonical.awk`.
- `dotseq` added to the `--translational_efficiency_method` enum; new `--extra_orf_dte_args` and `--extra_dotseq_args`.
- Pipeline test `tests/dotseq.nf.test`; docs (usage + output), CHANGELOG and CITATIONS.

## Considered and dropped

We prototyped an extra step that re-derived the gene-level Ribo-seq numerator by summing only the main-CDS ORF counts, to keep uORFs out of the gene-level TE. It's dropped: in the plastid path that numerator is already CDS-restricted (so already uORF-free), and re-deriving it could only ever zero genes whose main ORF wasn't confidently called (and short CDSs demoted to smORF), biasing the gene-level fit for no benefit. uORF dynamics are captured at ORF level (DOU) instead.

Closes #168


modules/local/dte_counts_prep/templates/dte_counts_prep.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+def load_mapping(path, primary_col, secondary_col):
+    df = pd.read_csv(path, sep="\\t")
+    for col in (primary_col, secondary_col):
+        if col not in df.columns:
+            raise SystemExit(f"{path}: expected `{col}` header")
+    df = df[[primary_col, secondary_col]].dropna()
+    df = df[(df[primary_col] != "") & (df[secondary_col] != "")]
+    df = df.drop_duplicates(subset=[primary_col], keep="first")
```
suhrig: I wonder if there is a biologically more meaningful method to select the RNA denominator when the same (collapsed) ORF is shared by multiple genes. How about the gene with the highest count? This gene should be the main driver of the ORF expression and, therefore, should be a better baseline IMO.

Well, the gene with the highest length-corrected count would actually be better, since the gene with the highest raw count will be biased towards longer genes. But we don't have the gene length here. So the highest raw count is the best we can do without making this process a whole lot more complex ...

nextflow_schema.json
```diff
@@ -502,14 +502,23 @@
...
-                    "enum": ["anota2seq", "deltate"],
+                    "description": "Method to use for translational efficiency analysis. `dotseq` is ORF-level only and requires `--extended_orf_analysis true`.",
+                    "enum": ["anota2seq", "deltate", "dotseq"],
                     "fa_icon": "fas fa-balance-scale"
                 },
                 "extra_deltate_args": {
                     "type": "string",
                     "description": "Extra arguments to pass to deltaTE (DESeq2) in addition to defaults defined by the pipeline."
                 },
+                "extra_orf_dte_args": {
```
suhrig: `extra_orf_deltate_args` would be a name that is more consistent with its gene-level counterpart (`extra_deltate_args`).

Speaking of: do we also need a `extra_orf_anota2seq_run_args`?

nextflow_schema.json
```diff
@@ -502,14 +502,23 @@
                 "translational_efficiency_method": {
                     "type": "string",
                     "default": "anota2seq",
-                    "description": "Method to use for translational efficiency analysis.",
-                    "enum": ["anota2seq", "deltate"],
+                    "description": "Method to use for translational efficiency analysis. `dotseq` is ORF-level only and requires `--extended_orf_analysis true`.",
+                    "enum": ["anota2seq", "deltate", "dotseq"],
```
suhrig: I may be asking for too much (let me know): Is there a reason why this parameter accepts only one value? Can't we run all tools in parallel? They are the final output of the pipeline, so there is no need to handle wiring of multiple tools to downstream processes. And the output files are disambiguated by being placed in different subfolders and harboring the name of the tool in the file name. => It shouldn't be too complicated to make this an argument that accepts multiple options.

There are two reasons I'm asking for this:
1. Convenience - enabling the user to generate everything they may possibly want in a single run.
2. When DOTseq is selected, the gene-level analysis is skipped, because DOTseq does not support it. A user would need to rerun the pipeline with `anota2seq` or `deltate` if they wanted to have the gene-level results as well.

suhrig (changes requested):

modules/local/dte_counts_prep/templates/dte_counts_prep.py
pinin4fjords: Good call, done. When an ORF maps to several genes it now takes the host gene present in the RNA matrix with the highest total count as the denominator, rather than an arbitrary first hit. Kept it to raw counts as you suggest (no gene length available here to correct for), with a comment flagging the mild length bias. The module test now covers it: an ORF mapping to two present genes picks the higher-count one over file order.

nextflow_schema.json
pinin4fjords: Agreed, renamed to `--extra_orf_deltate_args`. And good spot on the asymmetry: the ORF-level anota2seq was quietly reusing the gene-level `--extra_anota2seq_run_args`, unlike deltaTE which had its own. Added `--extra_orf_anota2seq_run_args` and pointed the ORF run at it, so both are independently tunable now.

nextflow_schema.json
pinin4fjords: Not too much at all, good idea. `--translational_efficiency_method` now takes a comma-separated list and each method runs independently (the outputs already land in per-tool subfolders, so there's no downstream wiring to worry about). That covers both your points: everything in a single run, and gene-level results are available even when dotseq is selected. Added `tests/te_multi_method.nf.test` running anota2seq + deltate together.

pinin4fjords:
Thanks Sebastian, all three addressed in the latest push: highest-count RNA denominator, arg renaming plus a matching `--extra_orf_anota2seq_run_args`, and multi-select for `--translational_efficiency_method`. Snapshots regenerating now.

suhrig (approved):
Excellent work, thank you! Time and again, I'm blown away by your output.

pinin4fjords:
> Excellent work, thank you! Time and again, I'm blown away by your output.

Thanks! It really is a new world with AI tooling.

nextflow_schema.json
bediagan: Thanks for addressing this! In my experience with human and mouse genomes, the vast majority of ORFs are unequivocally linked to a single gene on a specific strand anyway. However, for those rare multi-gene edge cases (like unstranded data, read-throughs, or paralogs), using the highest raw count to identify the dominant driver is a much better, more robust default than just grabbing an arbitrary first hit. Great addition!

As for the raw data, I believe keeping it as raw counts is mathematically necessary here. The downstream DTE engines rely on raw integer counts to correctly calculate library size and dispersion—either directly through negative binomial models (like deltaTE/DESeq2) or to perform proper internal variance-stabilizing transformations (like anota2seq). I believe, passing pre-normalized or length-scaled metrics like TPM into the matrix would break these statistical setups.

PR #194, opened 2026-07-03
chore: update all modules and subworkflows to latest

pinin4fjords:
## Description

Release-preparation update that brings every nf-core `modules/` and `subworkflows/` component up to its latest revision (via nf-core/tools 4.0.2) and reconciles the pipeline with the interface changes those updates introduce. Scope is deliberately limited to component updates and the wiring needed to accommodate them; no pipeline template sync.

## Components updated

- All 81 nf-core modules and 20 nf-core subworkflows updated to their latest `master` revisions (`nf-core modules update --all`, `nf-core subworkflows update --all`).
- Notable tool bumps captured in the snapshots: MultiQC 1.35, samtools 1.23.x, salmon 1.10.3, and the STAR/UMI/bedtools/plastid/ribo* toolchain.

## Interface reconciliation

The component bump carried breaking interface changes, accommodated in the local workflow/subworkflows:

- **samtools ecosystem**: the `bai` output is renamed to `index` (propagated through the STAR alignment and UMI-dedup subworkflow call sites); `samtools/view` gained a region-BED input, added in the coverage-tracks subworkflow.
- **fasta+fai inputs**: `fastq_align_star` and `bam_dedup_umi` now take `[meta, fasta, fai]` tuples. Reference channels are built as value channels (`.first()`) so the per-sample `BAM_SORT_STATS_SAMTOOLS` reference is reused across all samples.
- **quantify_pseudo_alignment**: pass the new `skip_merge` input.
- **utils_nfschema_plugin**: pass the new `cli_typecast` input.
- **topic-channel versions**: nf-core components migrated their `versions.yml` path outputs to the `versions` topic channel; the now-removed `.out.versions` emits are dropped where the topic channel already collects them (anota2seqrun, the quantify subworkflows), and the QC subworkflow's `multiqc_files` (now `[meta, file]` tuples) has its meta stripped before the MultiQC file collection.

## Dependencies

- **nf-schema 2.5.1 -> 2.7.2**, fixing boolean CLI parameter validation under the Nextflow 26.x strict-syntax parser.
- **Minimum Nextflow raised to 25.10.0** (required by nf-schema 2.7.2); CI matrix and README badge updated to match.

## Testing

All 11 pipeline nf-test snapshots regenerated on x86_64 under Nextflow 25.10.0 with nf-test 0.9.3, matching CI.


.nf-core.yml
```diff
@@ -1,4 +1,6 @@
 lint:
+  merge_markers:
+    - modules/nf-core/bowtie2/align/tests/main.nf.test.snap
```
luisas, line 3: I guess this one is a workaround for the << entries in the test right?

luisas (approved):
LGTM!

.nf-core.yml
pinin4fjords: Yep! Probably shouldn't be that way in module, but I don't have the will to fix it right now!

Thanks for review

Issue #202, opened 2026-07-18
Hybrid GTF wiring inconsistency between Rpbp and PRICE


suhrig:
### Description of the bug

When extended ORF analysis is activated, Rpbp receives the hybrid ORF, which means Rpbp can call ORFs from novel transcripts detected by Stringtie:

https://github.com/nf-core/riboseq/blob/68002f9b8da43389a23d8f9b4f6ac89dc9d95409/subworkflows/local/orf_caller_dispatch/main.nf#L147-L157

In contrast, PRICE always gets the multi-transcript GTF regardless of the extended ORF status, which means PRICE will never detect ORFs from novel transcripts:

https://github.com/nf-core/riboseq/blob/68002f9b8da43389a23d8f9b4f6ac89dc9d95409/subworkflows/local/orf_caller_dispatch/main.nf#L173-L182

Shouldn't PRICE be wired analogously to Rpbp?

### Command used and terminal output

```console

```

### System information

nf-core/riboseq dev version (https://github.com/nf-core/riboseq/commit/68002f9b8da43389a23d8f9b4f6ac89dc9d95409)

pinin4fjords:
Fixed by #209 (merged to `dev`). GitHub's auto-close via the "Closes #202" keyword didn't fire because the PR merged into `dev`, not the default `master` branch.

PR #204, opened 2026-07-20
Include novel-transcript ORFs in ORF-level DTE

pinin4fjords:
## Summary

The pipeline detects ORFs on novel StringTie transcripts and quantifies them, but ORF-level differential translation (DTE) was silently dropping every ORF on a novel gene. This gives novel genes an RNA-seq denominator so their ORFs are retained.

Raised by @suhrig on Slack

## Root cause

ORF-level DTE (`DTE_COUNTS_PREP`) joins per-ORF ribosome counts with a gene-level RNA-seq matrix and drops any ORF whose host gene is missing from that matrix. The RNA matrix was quantified against the canonical reference only, so novel StringTie genes had no row and their ORFs were excluded; ORFs on known genes survived.

## Fix

Quantify the denominator against the full reference transcriptome plus the novel intergenic transcripts, using the same STAR-align then Salmon path as the primary RNA-seq quantification (UMI dedup under `--with_umi`). Novel genes gain a row; canonical genes keep their full-isoform counts. The augmented annotation is the full reference GTF with the hybrid GTF's novel-gene rows appended (`assets/build_full_hybrid_gtf.awk`); outputs land under `quantification/salmon_hybrid/`.

## Key decisions

- **ORF-level DTE only.** Gene-level TE stays canonical-only, unchanged; that is exactly what was raised.
- **Full reference + novel, not the one-transcript-per-gene backbone** used for ORF calling: the backbone undercounts multi-isoform genes (16-65% in test data), which would shift every canonical ORF's result.
- **Same method as the primary (STAR-align then Salmon), not a lighter pseudo-alignment:** pseudo is a different method and gave a systematic ~2x offset; matching the primary keeps canonical genes consistent (identical library totals, only small bidirectional per-gene noise that normalises out in the contrast).
- **Only genuinely novel genes are added;** known-gene novel isoforms already have a denominator via their gene.
- Fixed `ch_hybrid_gtf` to a value channel (it was a single-item queue that starved a per-sample fan-out).

## Caveats

- These ORFs are hosted by a novel transcript, so they legitimately get a denominator; this is not the "no-host intergenic ORF, exclude from DTE" case. They inherit the existing ORF-level caveat already documented in `usage.md` (anota2seq/deltaTE assume one row per gene, so sibling-ORF p-values are a ranking, not strict significance; DOTSeq is the purpose-built ORF-level tool).
- **Test coverage:** the chr20 `-profile test` data contains no novel-transcript ORF (StringTie finds none; intergenic loci have no Ribo-seq signal), so the regenerated snapshots only confirm the pipeline runs, canonical DTE is preserved, and the `salmon_hybrid` outputs appear. Demonstrating a rescued novel ORF needs `test_full`-scale data.


suhrig (approved):
Okay from my side. Let's give @naiarabediaga a chance to review, too.

The approach of using the full reference + novel, not the one-transcript-per-gene backbone may also be a workable solution for issue https://github.com/nf-core/riboseq/issues/202.

Issue #197, opened 2026-07-14
Unused parameters


suhrig:
### Description of the bug

While running some tests, I noticed that the parameter `--min_mapped_reads` has no function and can be dropped from `nextflow_schema.json`/`nextflow.config`. Looks like a remnant from the rnaseq workflow to me.

I then searched systematically and found the following to have no function either:
- `fastp_merge`: bypassed, because hardcoded to `false` on line 192 in workflows/riboseq/main.nf. => Workflow should respect setting.
- `pseudo_aligner`: only used to check for existence of Salmon index (workflows/riboseq/main.nf:165), otherwise ignored and hardcoded to `salmon` (workflows/riboseq/main.nf:558 and workflows/riboseq/main.nf:587). => Either we drop support for kallisto or the workflow should respect the parameter.
- `skip_pseudo_alignment`: only has an effect on whether the Salmon index is created (workflows/riboseq/main.nf:165), doesn't actually gate the pseudo-alignment processes. When I set it to false, the pipeline crashes with `Salmon failed to produce lib_format_counts for sample 'SRX11780888' which was set to 'auto' strandedness. Check that the Salmon index matches your input reads, or set strandedness explicitly in the samplesheet`. I propose to drop this parameter, and `skip_alignment`, too. I don't think an "index only" run is a common use case. And now that the pipeline make use of so many tools that generate indices, the parameters would only be partially useful, even if they worked.

### System information

This is in reference to the version of the dev branch (https://github.com/nf-core/riboseq/commit/68002f9b8da43389a23d8f9b4f6ac89dc9d95409).

pinin4fjords:
Fixed by #212 (merged to `dev`). GitHub's auto-close via the "Closes #197" keyword didn't fire because the PR merged into `dev`, not the default `master` branch.

PR #206, opened 2026-07-27
refactor: streamline output folder structure (#196)

### LINKED ISSUE #196: Streamline output folder structure
### Description of feature

Here are some suggestions to streamline the output folder structure of the current dev branch (https://github.com/nf-core/riboseq/commit/68002f9b8da43389a23d8f9b4f6ac89dc9d95409). I consider them low-priority.

- PRICE, RiboCode, and RPBP spill some output files in the root of the output directory. They should go to the subfolder `orf_predictions`.
- ORF-level DTE results are written to `dte/orf_level/anota2seq`. They should be moved next to the gene-level results, i.e., `translational_efficiency/orf_level/anota2seq`.
- I think `hybrid_star` should be moved to `alignment/hybrid_star`. The folder `hybrid_star/index` should go to `genome/index/hybrid_star`.
- Optional: PRICE stores the index in the output files under `price/index`. Analogous to the STAR index, this index should only be saved, when `--save_reference` is set. (Unless there is a reason to keep the index which I am unaware of.) And would the `genome/index` output folder be a better place for the index, since all other indices go there as well?
- Optional: Shall we move `orf_cataloge` to `orf_predictions/catalogue`?
- Optional: Shall we move `orf_quantification` to `quantification/orf_level` to be consistent with `translational_efficiency/orf_level`?
- Optional: Shall we move `stringtie` to a subfolder named after its function, i.e., `transcript_assembly/stringtie`?

pinin4fjords:
## Summary

Streamlines the output folder structure described in #196. PRICE, RiboCode, and Rp-Bp outputs are grouped under `orf_predictions/`; ORF-level DTE results sit alongside gene-level results under `translational_efficiency/orf_level/`; and StringTie, ORF catalogue, quantification, and hybrid STAR outputs use the same category-based layout as the rest of the pipeline.

## Output changes

```
Before                                  After
------                                  -----
<outdir>/price/                         <outdir>/orf_predictions/price/
<outdir>/price/index/                   <outdir>/genome/index/price/
<outdir>/ribocode/                      <outdir>/orf_predictions/ribocode/
<outdir>/rpbp/                          <outdir>/orf_predictions/rpbp/
<outdir>/hybrid_star/                   <outdir>/alignment/hybrid_star/
<outdir>/hybrid_star/index/             <outdir>/genome/index/hybrid_star/
<outdir>/orf_catalogue/                 <outdir>/orf_predictions/catalogue/
<outdir>/orf_quantification/            <outdir>/quantification/orf_level/
<outdir>/stringtie/                     <outdir>/transcript_assembly/stringtie/
<outdir>/dte/orf_level/                 <outdir>/translational_efficiency/orf_level/
```

Reference-like outputs from PRICE and Rp-Bp genome preparation are published only with `--save_reference`. All outputs from the Rp-Bp subworkflow have explicit publishers under `orf_predictions/rpbp`, including metagene profiles, selected offsets, length-offset tables, profile matrices, Bayes factors, and final predictions.

The output documentation and nf-test snapshots are updated to cover the new paths. The `dotseq`, `novel_gtf`, and `stringtie_extended` tests assert the full hybrid STAR output tree instead of excluding it.

Closes #196


FelixKrueger (commented):
Nice cleanup — the tree makes a lot more sense this way, and the docs/schema updates are consistent all the way through. Two things I'd like sorted before approving, plus two minor points.

## 1. The root-level `rpbp/` spill is still there

The fallback publisher builds its path from the process-name prefix:

```groovy
path: { "${params.outdir}/${task.process.tokenize(':')[-1].tokenize('_')[0].toLowerCase()}" }
```

so *any* uncovered `RPBP_*` process lands in `<outdir>/rpbp/`. Rp-Bp has eight processes and four are covered (`PREPAREGENOME`, `ESTIMATEMETAGENEBAYESFACTORS`, `ESTIMATEORFBAYESFACTORS`, `SELECTFINALPREDICTIONSET`). These four still fall through:

- `RPBP_EXTRACTMETAGENEPROFILES` → `<outdir>/rpbp/*.metagene.csv.gz`
- `RPBP_SELECTPERIODICOFFSETS` → `<outdir>/rpbp/*.offsets.csv.gz`
- `RPBP_GETPERIODICLENGTHSOFFSETS` → `<outdir>/rpbp/*.lengths-offsets.tsv`
- `RPBP_EXTRACTORFPROFILES` → `<outdir>/rpbp/*.profiles.mtx.gz`

All four are called unconditionally in `subworkflows/nf-core/fasta_gtf_bam_rpbp/main.nf`, so with `--run_rpbp true` you still get a bare `rpbp/` at the output root next to `orf_predictions/rpbp/`. `run_rpbp` defaults to `false`, which is presumably why no test catches it. RiboCode is fully covered by comparison — all four of its processes.

A subworkflow-wide block would catch all of them and stay correct if Rp-Bp gains steps later:

```groovy
withName: '.*:FASTA_GTF_BAM_RPBP:.*' {
    publishDir = [
        path: { "${params.outdir}/orf_predictions/rpbp" },
        mode: params.publish_dir_mode,
        saveAs: { filename -> filename.equals('versions.yml') ? null : filename }
    ]
}
```

Keeping the three existing blocks after it so their `ext.args` / `time` / `pattern` still win.

## 2. Three tests quietly stop asserting the `hybrid_star` tree

`dotseq`, `novel_gtf` and `stringtie_extended` each gained `'alignment/hybrid_star/**'` in the `stable_name` ignore list. That one isn't a rename — it drops 101 filename entries per snapshot (`log` 25, `original` 13, `sorted` 62, `unmapped` 1, `index` 1), leaving only the bare directory entry.

The effect is that the name-level assertion no longer covers the tree this PR relocates: if the hybrid pass published nothing at all, or published under a wrong sub-path, all three tests would still pass. It also means "each diff is path-only with every content hash unchanged" and "a misplaced file would show up as a diff" don't hold for `hybrid_star` any more.

Current `dev` asserts all 101 of those names and is green, so they look stable — was this working around something specific? If so, a narrower pattern for just the unstable subdirectory (plus a line in `.nftignore_orf` saying why) would keep the rest of the coverage.

## Minor

- `RPBP_PREPAREGENOME` publishes its three annotated BEDs unconditionally, while the PRICE index is now `--save_reference`-gated. Same class of reference-prep artefact, different treatment — deliberate?
- The conflict note (#213 should land first) and "snapshots need regenerating once more on the merge" are both stale: #213 is merged, the branch isn't behind `dev`, and CI is green on 535ed1f8. Worth trimming so the merge commit doesn't carry them.

## Checked and fine

No old paths left anywhere in `conf/`, `docs/`, `tests/` or the subworkflows; the base `.nftignore` genuinely needed no change; the PRICE index gating is safe (`GEDI_PRICE` receives `GEDI_INDEXGENOME.out.index` on a channel, so it stages from the work directory, never the publish path); and #213's anchored selectors survived the merge intact.

FelixKrueger (commented):
The `--save_reference` gate on `RPBP_PREPAREGENOME` (`conf/modules.config:1504`) doesn't take effect: the `.*:FASTA_GTF_BAM_RPBP:.*` catch-all at `:1496` overrides it, because a `:`-bearing `withName` selector outranks a bare-name one regardless of declaration order. The prepare-genome BEDs therefore still publish unconditionally, while `docs/output.md:459` now tells users they only appear with `--save_reference`. Not a regression — base published them unconditionally too — but the config reads as though the gate is live, and `run_rpbp` defaults false so no CI job exercises it. One-line fix is to qualify the selector as `.*:FASTA_GTF_BAM_RPBP:RPBP_PREPAREGENOME`. The same shadowing makes `:1512`'s `publishDir` dead code, though its `time` directive is unaffected since the override is per-directive rather than whole-block.

Three smaller things: the CHANGELOG entry reads as housekeeping but renames five user-visible output directories, so it could use a **Breaking:** prefix like #197's; `nextflow_schema.json:654` still describes `--save_reference` as "the STAR index" although this PR extends it to the PRICE index; and `RIBOTRICER_PREPAREORFS` (`:1012`) is still on the implicit default, so `--run_ribotricer true` drops a stray top-level `ribotricer/` — the same pattern you fixed here for RiboCode and Rp-Bp.

The path move itself looks correct, including the snapshot re-records and the `.nftignore_orf` repointing. Details below.

<details>
<summary>Verification notes (AI-assisted review)</summary>

Two independent reviews plus a tie-break experiment, since the reviewers initially disagreed about whether the gate applied.

**Selector precedence, tested rather than assumed.** Nextflow 25.10.4, minimal pipeline mirroring the real selector shapes and subworkflow nesting, catch-all declared *first* and ungated, gated bare-name second, `save_reference = false`:

```
Submitted process > RIBOSEQ:FASTA_GTF_BAM_RPBP:RPBP_PREPAREGENOME
Submitted process > RIBOSEQ:FASTA_GTF_BAM_RPBP:RPBP_SELECTFINALPREDICTIONSET

res2/CATCHALL/prepared.bed      <- ungated catch-all won, save_reference=false
res2/QUALIFIED/final.bed        <- fully-qualified selector won (control)
                                <- no GATED/ directory was created
```

The rule appears to be two-tier: across specificity classes a `:`-bearing selector wins regardless of order, while within one class the later declaration wins. That second half is why `SEQKIT_REPLACE`'s `:478` block correctly shadows `:438` — both are bare names. Worth a comment at `:1496`, since the distinction is easy to trip over.

**Snapshots — verified a pure path remap.** `stable_path` is byte-identical as a multiset between base and head in all four tests, so no md5 changed and nothing was added or dropped:

```
dotseq                         base=315 head=315  only-base=0 only-head=0
novel_gtf                      base=316 head=316  only-base=0 only-head=0
orf_catalogue_no_novel_source  base=297 head=297  only-base=0 only-head=0
stringtie_extended             base=315 head=315  only-base=0 only-head=0
```

That check is bidirectional, which is the point: a pattern that stopped matching would *add* an entry, one that over-matched would *remove* one. The entry lists are reordered (~92 of ~315 positions), consistent with a genuine re-record rather than hand-editing. `orf_catalogue_no_novel_source`'s net −1 in `stable_name` is directory bookkeeping — base had both `dte/` and `translational_efficiency/` at top level, head has only the latter, and it is the one test where StringTie doesn't run so it gains no `transcript_assembly/` entry. No file entry lost.

**`.nftignore_orf` reconciles exactly.** Re-deriving each snapshot's `stable_path` from its `stable_name` by applying all 113 patterns with Java-glob semantics reproduces the recorded list exactly in all four tests (residuals are the expected empty `publishDir` targets). Applying the *base* ignore file to *head* paths as a sensitivity control would have leaked 18–130 extra files per test, so the check isn't vacuous. All 12 repointed patterns match real files at their new locations; `quantification/salmon_hybrid/tx2gene.tsv` is still correctly suppressed while the primary `quantification/tx2gene.tsv` is hashed. `tests/.nftignore` correctly needed no change.

**Also checked clean:** no stale old path anywhere in code, config, docs, schema, assets or the other eight snapshots; the `+58/−30` in `conf/modules.config` is six new explicit blocks replacing implicit defaults, with no directory collisions and no `saveAs`/`pattern`/`enabled` condition lost; `docs/output.md`'s tree matches what the config produces; and gating the PRICE index is safe because `GEDI_PRICE` receives it as a staged channel input rather than reading the published copy.

**Minor, take or leave:** `tests/.nftignore_orf` has no `translational_efficiency/orf_level/deltate/**` alongside `anota2seq` and `dotseq` (pre-existing, unreachable today, but the three now sit side by side); `tests/.nftignore:70`'s `translational_efficiency/*/*.{...,tsv}` glob now also covers `orf_level/orf_combined_counts.tsv`, harmless while every ORF-level test uses the ORF ignore file; and `docs/output.md:857`'s `genome/index/` bullet doesn't list the new `hybrid_star/` and `price/` children.

Worth noting for sequencing: this PR changes no content md5, so landing it before the in-flight `orf_class` vocabulary work means that branch's re-record attributes every md5 change to the vocabulary alone. The reverse order is worse — those `.snap` files are JSON lists and wouldn't conflict, so a stale-path recording would auto-merge silently.

</details>

FelixKrueger (approved):
Thanks — everything from my earlier review is addressed, and `14e6888` fixes the selector correctly. Both selectors now carry `:`, so they're in the same specificity class and `.*:FASTA_GTF_BAM_RPBP:RPBP_PREPAREGENOME` at `:1508` wins on declaration order over the catch-all at `:1500`; the `--save_reference` gate is live. Dropping the redundant `publishDir` from the `RPBP_ESTIMATE*` block while keeping its 30 h `time` is right too, and `RIBOTRICER_PREPAREORFS` now has an explicit path with the docs to match.

CI is 34/34 green on `14e6888`, `confirm-pass` included, and the `pr-comment` artifact carries no `comment.md`, so no `latest-everything` leg failed either.

One note for sequencing rather than a request: this PR changes no content md5, so landing it before the `orf_class` vocabulary work keeps the two sources of snapshot churn separate — that branch's re-record then attributes every md5 change to the vocabulary alone.

PR #185, opened 2026-05-22
feat: PRICE opt-in ORF caller

pinin4fjords:
## Summary

Adds [PRICE](https://github.com/erhard-lab/gedi/wiki/Price) (Erhard et al. 2018) as an opt-in ORF caller via `--run_price` (default `false`). PRICE estimates a shared cohort-level codon-position model via EM and is invoked one-shot across the riboseq cohort (`gedi -e IndexGenome` then `gedi -e Price`), not per-sample.

## Changes

- Install nf-core gedi modules: `gedi/indexgenome`, `gedi/price` (nf-core/modules#11693, merged). Container ships via the merged `bioconda::gedi=1.0.6a` recipe.
- New params: `--run_price`, `--extra_price_indexgenome_args`, `--extra_price_price_args`.
- Workflow integration: PRICE block alongside the existing ORF callers, honours `--extended_orf_analysis` (IndexGenome consumes the hybrid GTF when extended-ORF mode is active).
- conf/modules.config: `withName` blocks for GEDI_INDEXGENOME (prefix `price_index`) and GEDI_PRICE, publishing under `<outdir>/price/`.
- docs + CHANGELOG.

## Caveats

- PRICE is validated at the module level (nf-core/modules `gedi/price`, chr19+22 curated cohort, ~380 ORFs); there is no pipeline-level test. PRICE's `PriceOrfInference` fails on STAR alignments at `-profile test` genome scale (chr20, and even chr19+22) regardless of read depth, while the same reads succeed as the module's curated full-genome-derived BAMs. End-to-end PRICE is therefore exercised at full scale on the Platform iteration.
- Runtime is comparable to other heavy callers at genome-wide scale; the pipeline emits a runtime warning when `--run_price` is set.

## Stacked PR notes

Ninth in the stack splitting #174. Targets #184 (`feat/171-second-star-pass`).

Closes #170

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #207, opened 2026-07-27
fix: resolve PRICE ORF gene ids from the annotation the callers actually ran against

### LINKED ISSUE #203: Underscores in gene IDs in `orf_catalogue/cohort.catalogue.tsv`
### Description of the bug

A number of entries in `orf_catalogue/cohort.catalogue.tsv` contain an underscore in the `gene_id` column. This is because PRICE reports concatenated gene IDs separated by an underscore when an ORF overlaps with multiple genes. PRICE is the only tool to do this and this behavior causes a bit of confusion when reading the ORF catalog. It will also lead to ORFs **that are only called by PRICE** to be missing from `dte/orf_level/orf_combined_counts.tsv`, because this table is joined with the RNA-seq count matrix from Salmon, which only contains regular gene IDs.

A better approach would be to resolve the gene IDs from the transcript IDs that PRICE reports in its output file `price/allsamples.orfs.tsv` in the column `Id` and ignore the gene IDs stated in the column `Gene`. The transcript ID provides a unique mapping to the gene.

```console
nextflow run riboseq/main.nf -profile conda,test_full --outdir=$PWD/results --run_price true
```

### Relevant files

_No response_

### System information

### LINKED ISSUE #219: PRICE gene_id resolution only partially fixed: orfnormalise gets the canonical backbone, PRICE runs against the full annotation
Follow-up from reviewing #207, which fixed the PRICE `gene_id` resolution inside `custom/orfnormalise` (nf-core/modules#12436). That fix is correct in the module, but it can only take effect for ORFs whose transcript is present in the annotation `orfnormalise` is handed — and in this pipeline that annotation is narrower than the one PRICE ran against. So #203 is only partly closed.

## The mismatch

PRICE and `orfnormalise` receive different GTFs:

- `subworkflows/local/orf_caller_dispatch/main.nf:180` — PRICE gets `ch_gtf`, the full multi-isoform annotation. This is deliberate, and the in-code comment explains why: restricting PRICE "to the canonical/hybrid annotation would only narrow ORF discovery and bias ORF-type classification to canonical CDS".
- `workflows/riboseq/main.nf:477` — `ORFTABLE_FASTA_GTF_BUILDORFCATALOGUE`, and therefore `CUSTOM_ORFNORMALISE`, gets `ch_hybrid_gtf`. That resolves to `ch_canonical_gtf`, the one-transcript-per-gene backbone (longest isoform per gene via gffread when `--canonical_gtf` isn't supplied — `subworkflows/local/prepare_genome/main.nf:140-159`), plus filtered novel intergenic transcripts when novel discovery is active.

## Consequence

After #207, `parse_price` resolves `gene_id` by looking the ORF's transcript up in the supplied annotation, and falls back to PRICE's `Gene` column only when that transcript is absent. A PRICE ORF sitting on any isoform that isn't the per-gene representative has a transcript id that isn't in the backbone, so:

1. `transcripts.get(tid)` misses,
2. it falls back to the `Gene` column, which for a multi-gene-overlapping ORF is the underscore-concatenated compound value,
3. the compound id matches nothing in the Salmon gene-level matrix, and the ORF drops out of `dte/orf_level/orf_combined_counts.tsv`.

That is exactly the #203 symptom, surviving for the subset of affected ORFs that sit on non-representative isoforms. Because PRICE is deliberately run against the full multi-isoform annotation, I'd expect that subset to be a meaningful fraction of PRICE-only ORFs rather than a rare tail — though I haven't been able to quantify it, since a `--run_price true` run isn't currently feasible (see #185).

Note the fallback is not a regression — it is precisely the pre-#207 behaviour — so this is "the fix doesn't reach far enough", not "the fix broke something".

## Candidate fix

`ch_full_hybrid_gtf` already exists for this purpose. Built at `workflows/riboseq/main.nf:698-702`, it is the union of the full multi-isoform annotation and the hybrid novel transcripts, and it is what the RNA-seq quantification reference is built from (added in #204). That makes it the natural annotation for ORF gene resolution: it is a superset of what PRICE saw, and its gene namespace is by construction the one the DTE join has to match.

Things to work through before doing it:

- **Blast radius.** Switching the catalogue's GTF changes `gene_id` (and for ribocode/ribotish, coordinate lifting) for *every* caller, not just PRICE. Other callers currently resolve against the backbone on purpose. Plausibly also an improvement, but it will move existing outputs and snapshots, so it needs its own PR and its own justification per caller.
- **Ordering.** `ch_full_hybrid_gtf` is constructed inside the DTE branch, well after the catalogue is built. Feeding it to the catalogue means hoisting its construction, or building it unconditionally.
- **Availability.** It only exists when the relevant RNA-seq/DTE path is active, so the catalogue would need a sensible fallback to `ch_hybrid_gtf` when it isn't.
- **Alternative, narrower option.** Pass a second, price-specific annotation to the normaliser rather than changing the shared one, keeping the other callers untouched. Costs an interface change to the nf-core subworkflow and module, so heavier upstream but zero collateral movement here.

## Also stale, and related

`custom/orfnormalise`'s consumers still document the GTF as irrelevant to PRICE, which is no longer true now that it is PRICE's primary gene source:

- `subworkflows/nf-core/orftable_fasta_gtf_buildorfcatalogue/meta.yml:59-65` — "ignored by callers that already emit genomic coordinates (ribotricer, rpbp, price)", and the claim that `[ [:], [] ]` is acceptable provided the caller subset excludes ribocode and ribotish. With #207 in place, passing an empty GTF silently degrades every PRICE row to the compound `Gene` value.
- The equivalent comment in that subworkflow's `main.nf:23-24`.

Both live in an nf-core subworkflow, so correcting them needs an upstream nf-core/modules PR rather than a change here.

## Suggested verification

Whatever route is taken, the check that actually demonstrates closure is a module-level `custom/orfnormalise` test where the PRICE fixture's transcript is a non-representative isoform absent from the supplied GTF, asserting the emitted `gene_id` is still the plain one. The regression test added in #207 uses a transcript that *is* present, so it cannot catch this class.

pinin4fjords:
Closes #203, closes #219. Thanks to @suhrig for the report and @FelixKrueger for spotting the pipeline-level half.

## Problem

PRICE-only ORFs carried a compound `gene_id` in `orf_catalogue/cohort.catalogue.tsv` that never matched the plain `gene_id` in the Salmon count matrix, dropping them out of `dte/orf_level/orf_combined_counts.tsv`. Two things caused it:

1. `parse_price` preferred PRICE's `Gene` column, which concatenates every gene an ORF's genomic span overlaps, over the GTF-derived transcript-to-gene mapping.
2. The catalogue was handed `ch_hybrid_gtf`, the one-transcript-per-gene backbone, while PRICE is deliberately run against the full multi-isoform annotation (`subworkflows/local/orf_caller_dispatch/main.nf`, so ORF discovery and ORF-type classification are not narrowed to canonical CDS). An ORF called on a non-representative isoform therefore had no transcript to resolve against, and fell back to the `Gene` column regardless of precedence.

## Fix

**Module (`custom/orfnormalise`, bumped to nf-core/modules#12436).** `gene_id` resolves from the transcript ID in the `Id` column via the supplied annotation, falling back to the `Gene` column only when that transcript is absent. In that fallback the raw value passes through unchanged, since there is no reliable way to pick one gene without a transcript match, and the ORF then fails to join exactly as any other unresolvable `gene_id` would. The `Id` parse is anchored on the row's own `orf_type` value so transcript IDs containing underscores survive, which `Id.split("_", 1)` did not handle. PRICE builds `Id` as `<transcript_id>_<orf_type>_<n>`, confirmed against the [erhard-lab/gedi wiki](https://github.com/erhard-lab/gedi/wiki/Price) and the real `cohort.price.orfs.tsv` fixture.

**Pipeline.** The catalogue now resolves against the union of the full reference annotation and the novel transcripts (`BUILD_FULL_HYBRID_GTF`), built once ahead of the caller dispatch and shared with the ORF-level DTE RNA denominator. That union is a superset of every annotation any caller is given, so no caller loses resolution, and it is by construction the gene namespace the DTE join has to match.

The union is deduplicated on `transcript_id` rather than `gene_id`. Keying on gene dropped novel isoforms of known genes, which the hybrid carries whenever `--gffcompare_class_codes` admits anything beyond intergenic (`u`) transcripts, both from the catalogue's annotation and from the RNA denominator. Output is byte-identical for the intergenic-only default.

## Effect per caller

| Caller | Annotation it runs against | `gene_id` source | Effect |
| --- | --- | --- | --- |
| PRICE | full multi-isoform | GTF via `Id` transcript | fixed: transcripts now present |
| Rp-Bp | hybrid (extended mode) | GTF via transcript id | unchanged, already resolvable |
| Ribo-TISH | hybrid + canonical | `Gid` column | unchanged |
| RiboCode | hybrid | own `gene_id` column | unchanged |
| Ribotricer | hybrid | own output | unchanged |

Coordinate lifting is unaffected: a transcript present in both annotations has the same exon blocks in either.

## Downstream consumers

None need changes. `orfmerge.py` and `orfcollapse.py` treat `gene_id` as an opaque string, `build_orf_count_matrix.py` keys only on `orf_id`, and `dte_counts_prep.py` already tolerates an ORF with several candidate genes, picking whichever is present in the RNA-seq matrix.

## Tests

Module-level: a PRICE row whose `Gene` holds a concatenated ID and whose underscore-containing transcript is present in a small GTF fixture, asserting the emitted `gene_id` is the plain one from the GTF. Ships with the module upstream, using generated inputs rather than committed fixtures.

Pipeline-level: `tests/stringtie_extended.nf.test` asserts every catalogue gene id exists in the RNA-seq matrix the ORF-level DTE join uses. This is a guard on the invariant, not a demonstration of the PRICE fix, which no pipeline test can reach: PRICE is opt-in and crashes on the pipeline's own STAR output regardless of depth (#185).

The extended-ORF snapshots are expected to be unchanged. Ribo-TISH and RiboCode take their gene ids from their own output columns, and the union contains every transcript the hybrid did, so nothing they emit moves.

## Interaction with #209

#209 gives PRICE and Rp-Bp the same union annotation and hoists the same build. The two are complementary (that one fixes what PRICE sees, this one fixes what the catalogue resolves against) and independent; whichever merges second drops the duplicated `BUILD_FULL_HYBRID_GTF` block.

## Not verified

`nf-core pipelines lint` was not run locally; CI covers it. `nextflow lint` reports only the pre-existing `utils_nextflow_pipeline` error.


FelixKrueger (approved):
Approving. The precedence inversion is the right call, and I convinced myself there's no regression path: enumerating old vs. new across all four branches, the only case where they differ is "transcript present in the annotation", and there the GTF-derived id is the correct namespace for the Salmon join. Every other case — transcript absent, `Gene` empty, both — is byte-identical to the old behaviour.

Things I checked:

- The vendored module is byte-identical to `nf-core/modules@c13f18d` across `main.nf`, `meta.yml`, `environment.yml` and `templates/orfnormalise.py`, so no patch file is needed, and `c13f18d` really is the merge commit of nf-core/modules#12436.
- Module tests ran upstream (9 shards × conda/docker/singularity, plus `nf-core lint modules (custom/orfnormalise)`), which is the right place given `nf-test.config` here sets `ignore = ['modules/nf-core/**/tests/*']`.
- The three downstream claims hold: `orfcollapse.py` never references `gene_id`, `build_orf_count_matrix.py` pivots on `orf_id` alone, and `orfmerge.py:267` uses `gene_id` only as an opaque dedup key.

## Follow-up: the fix is partial at pipeline level

Not a blocker, but worth an issue, because I don't think #203 fully closes with this.

PRICE and `orfnormalise` are handed different annotations:

- `subworkflows/local/orf_caller_dispatch/main.nf:180` — PRICE gets `ch_gtf`, the full multi-isoform annotation, deliberately so ("restricting it to the canonical/hybrid annotation would only narrow ORF discovery").
- `workflows/riboseq/main.nf:477` — the catalogue subworkflow, and so `orfnormalise`, gets `ch_hybrid_gtf`, which is `ch_canonical_gtf`: a one-transcript-per-gene backbone (longest isoform per gene via gffread when not supplied, `subworkflows/local/prepare_genome/main.nf:140-159`).

So a PRICE ORF sitting on any non-longest isoform has a transcript id that isn't in the GTF `orfnormalise` sees. `transcripts.get(tid)` misses, it falls back to the concatenated `Gene` column, and the ORF is still dropped from the join — the original symptom, surviving for the subset of affected ORFs on non-canonical isoforms. Since PRICE is deliberately run against the full annotation, I'd expect that subset to be a substantial fraction rather than a rare tail.

The candidate fix already exists in the pipeline: `ch_full_hybrid_gtf` (`workflows/riboseq/main.nf:698-702`) is the union of the full annotation and the novel transcripts, and it's exactly what the RNA-seq quantification reference is built from in #204 — i.e. the annotation whose gene namespace the DTE join has to match, and a superset of what PRICE saw. Passing that to the catalogue would close the gap, but it would also shift `gene_id` resolution for ribotish/ribocode, so it clearly belongs in its own PR rather than bolted onto this one.

## Minor

1. Contract docs are now stale: `subworkflows/nf-core/orftable_fasta_gtf_buildorfcatalogue/meta.yml:59-65` still says the GTF is "ignored by callers that already emit genomic coordinates (ribotricer, rpbp, price)" and that `[ [:], [] ]` is fine as long as the caller subset excludes ribocode and ribotish. Neither holds for PRICE any more — the GTF is now its *primary* gene source, and passing an empty one silently degrades every PRICE row to the compound `Gene` value. Same claim in that subworkflow's `main.nf:23-24`. Needs another upstream PR, so fair to defer, but worth tracking.

2. The new test doesn't discriminate the code path it was added for. With `Id = ENST_FAKE_TX_CDS_1` and `Type = CDS`, the `orf_type`-anchored regex and the `rsplit("_", 2)` fallback both return `ENST_FAKE_TX`, so the test still passes if the regex branch never matches. A discriminating case needs an `orf_type` that itself contains an underscore, and none of `PRICE_EXACT`'s keys do (`CDS`, `Ext`, `Trunc`, `Variant`, `uORF`, `uoORF`, `dORF`, `ncRNA`), nor do `iORF`/`intronic`/`orphan`. No bug here — the `rsplit` fallback is what actually delivers the underscore-safety, and it does so for any transcript id — but the regex is effectively belt-and-braces with no reachable behavioural difference on real PRICE output, which is worth knowing if it ever needs maintaining.

3. The CHANGELOG entry links the issue (`.../issues/203`) where every neighbouring entry links its PR. Numeric ordering is right. Purely cosmetic.

4. Unrelated to this PR, but it surfaced while reviewing it: changing a module's `templates/*.py` triggers no pipeline-level nf-test here, because nf-test's `--changed-since` graph follows `.nf` includes and not template files. This bump therefore got zero pipeline-level validation. Harmless in this instance since only `parse_price` changed and pipeline tests don't run PRICE, but it's a blind spot for template edits in general.

PR #208, opened 2026-07-27
fix: honour genome stanza index attributes (salmon, sortmerna, bbsplit)

### LINKED ISSUE #195: genome parameter for Salmon index not respected
### Description of the bug

When the `salmon_index` parameter is specified by setting the `salmon` attribute of a genome stanza, the pipeline runs the process `NFCORE_RIBOSEQ:RIBOSEQ:FASTQ_QC_TRIM_FILTER_SETSTRANDEDNESS:FASTQ_SUBSAMPLE_FQ_SALMON:SALMON_INDEX (genome.transcripts.fa)` nonetheless. Instead, it should make use of the existing index. When the Salmon index is passed via the command-line parameter `--salmon_index`, the pipeline does make use of it as expected.

```console
1. Add a genome stanza as follows to `conf/test.config`:


params {
    config_profile_name        = 'Test profile'
    config_profile_description = 'Minimal test dataset to check pipeline function'

    // Input data
    input = 'https://raw.githubusercontent.com/nf-core/test-datasets/riboseq/samplesheet/samplesheet.csv'
    contrasts = 'https://raw.githubusercontent.com/nf-core/test-datasets/riboseq/contrasts.csv'
    ribo_database_manifest = 'https://raw.githubusercontent.com/nf-core/test-datasets/riboseq/testdata/rrna-db.txt'
    genomes {
        'GRCh38chrom20' {
            fasta = 'https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/genomics/homo_sapiens/riboseq_expression/Homo_sapiens.GRCh38.dna.chromosome.20.fa.gz'
            gtf = 'https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/genomics/homo_sapiens/riboseq_expression/Homo_sapiens.GRCh38.111_chr20.gtf'
            salmon = '/home/sebastian/Desktop/riboseq/salmon/'
            star = '/home/sebastian/Desktop/riboseq/star/'
        }
    }
    min_trimmed_reads = 1000

    skip_ribocode = true
    extra_fqlint_args = '--disable-validator P001 --disable-validator S007'
}


2. Run the test:


nextflow run riboseq/main.nf -profile conda,test --outdir=$PWD/results --genome GRCh38chrom20


3. Note how `NFCORE_RIBOSEQ:RIBOSEQ:FASTQ_QC_TRIM_FILTER_SETSTRANDEDNESS:FASTQ_SUBSAMPLE_FQ_SALMON:SALMON_INDEX (genome.transcripts.fa)` is executed.
```

### Relevant files

_No response_

### System information

pinin4fjords:
Closes #195. Thanks to @suhrig for the report and the reproducer.

## Problem

Reference paths supplied through a `--genome` stanza were partly ignored. The reported symptom was a Salmon index being rebuilt in `FASTQ_SUBSAMPLE_FQ_SALMON:SALMON_INDEX` despite the stanza carrying a `salmon` attribute, where the equivalent `--salmon_index` worked.

`main.nf` resolves stanza attributes with script-level `params.x = getGenomeAttribute(...)` assignments. Two separate Nextflow rules were defeating that:

1. **A config default shadows a script-level assignment.** Nextflow's params map is read-only for keys arriving from configuration. `nextflow.config` declared `salmon_index = null` and `sortmerna_index = null`, so those two assignments were silently discarded everywhere, `main.nf` included.
2. **An included script snapshots `params` at its `include` statement.** The assignment block sat below the `include` lines, so `workflows/riboseq/main.nf` and everything it includes saw `null` for all nine reference params regardless of the stanza.

## Attribute audit

| Stanza attribute | Param | Read outside `main.nf` | Behaviour before this PR |
| --- | --- | --- | --- |
| `fasta`, `gtf`, `gff`, `transcript_fasta`, `additional_fasta` | respective | no | correct |
| `star` | `--star_index` | no | correct, the index is genuinely reused |
| `bbsplit` | `--bbsplit_index` | yes (rule 2) | index reused, but the BBSplit fasta-list check ran as if none was given |
| `salmon` | `--salmon_index` | yes (rules 1+2) | ignored, rebuilt for strandedness inference and in `PREPARE_GENOME` |
| `sortmerna` | `--sortmerna_index` | yes (rules 1+2) | ignored, rebuilt in both `PREPARE_GENOME` and `FASTQ_REMOVE_RRNA` |

## Fix

- `main.nf`: the genome attribute block resolves before `RIBOSEQ` and `PREPARE_GENOME` are included, so every consumer inherits the resolved values.
- `nextflow.config`: the `salmon_index` and `sortmerna_index` defaults are dropped so resolution is no longer shadowed. The other seven were already absent, and a comment records why they must stay that way.
- `docs/usage.md`: a "Custom genome stanzas" section listing the recognised attributes, their parameter equivalents and the precedence rule.

Explicitly set parameters (command line, `-params-file`, config `params` block) still take precedence over the stanza.

## Verified

With a stanza pointing at throwaway index directories, `-profile test,docker --genome REPRO195 -stub`: before, `FASTQ_SUBSAMPLE_FQ_SALMON:SALMON_INDEX` is created, `SORTMERNA_INDEX` is submitted in both `PREPARE_GENOME` and `FASTQ_REMOVE_RRNA`, and the run logs `Access to undefined parameter star_index` and `bbsplit_index`. After, none of those processes are created and the warnings are gone. `nextflow lint` is clean on both edited files.

Worth knowing: stanza-supplied `salmon_index` and `sortmerna_index` are now subject to the schema's `exists: true` check, so a stanza pointing at a missing directory fails validation instead of silently rebuilding.

There is no automated regression test. Reproducing this needs a stanza pointing at index directories that exist, which would mean carrying fixture directories for a case CI does not otherwise cover.

## Overlap with #197

#197 removes `pseudo_aligner` and `skip_pseudo_alignment` and rewrites the line whose `params.salmon_index` read is the visible symptom here. This PR does not touch that file, so the two should not conflict textually, and the fix holds either way because it makes `params.salmon_index` correct wherever it is read.


FelixKrueger (approved):
Approving. I reproduced both Nextflow rules this rests on in a scratch pipeline rather than taking them on trust, and they behave exactly as described:

| Scenario | `params.salmon_index` resolves to | Confirms |
| --- | --- | --- |
| script included **above** the assignment block | `<UNDEFINED>` | rule 2, include-time snapshot |
| script included **below** the assignment block | `RESOLVED_FROM_STANZA` | rule 2, and that the reordering fixes it |
| `salmon_index = null` present in `nextflow.config` | `null` **everywhere**, entry script included | rule 1, silent shadowing |
| explicit `--salmon_index` on the command line | `/explicit/cli/path` | the documented precedence |
| `paramsSummaryMap` called from an early-included subworkflow | `RESOLVED_FROM_STANZA` | summary and validation are unaffected by include order |

Worth recording the last row, because it is the one thing I expected to be a wart and isn't: `paramsSummaryMap` and `validateParameters` are plugin functions reading session params rather than the including script's snapshot, so they see stanza-resolved values even though `utils_nfcore_riboseq_pipeline` is necessarily included above the assignment block. The run summary stays honest, and your note about the two params now being subject to the schema's `exists: true` is right for the same reason.

I also like that rule 1 is what supplies the override semantics. Because config and CLI values are immutable, `--salmon_index` beats a stanza attribute for free — the same mechanism that caused the bug gives you the precedence guarantee, with no extra logic to maintain.

Static checks that back the audit table:

- All nine stanza params are absent from `nextflow.config`, not just the two you removed.
- Nothing included above the assignment block reads any of the nine, including the transitive `utils_nfcore_pipeline` / `utils_nextflow_pipeline` chain, so the ordering is safe.
- None of the nine carry a `default` in `nextflow_schema.json`, so dropping the two config defaults leaves nothing for lint to flag — borne out by the lint comment at 297/7/6, identical to the baseline on #207.
- No nf-test instantiates `workflows/riboseq` or `prepare_genome` directly, so removing the config defaults cannot produce undefined-parameter errors under test.
- `conf/igenomes.config` defines `star` across its stanzas but no `salmon`, `sortmerna`, `bbsplit` or `transcript_fasta`, so this changes nothing for genuine iGenomes keys and only affects custom stanzas — exactly the scope the new docs section claims.

None of what follows is blocking.

## The invariant is load-bearing, silent, and unguarded

The fix depends on nine parameters never being declared in `nextflow.config`, and nothing enforces that. Re-adding `salmon_index = null` is an inviting change for a future contributor, because the parameter *is* in `nextflow_schema.json`, so its absence from the config reads as an oversight rather than a deliberate invariant. By rule 1 the resulting breakage is completely silent — no warning, no error, just a rebuilt index.

CI cannot catch a regression either: `conf/test.config` sets `fasta` and `gtf` directly and no test uses `--genome`, so the stanza path is never exercised.

Cheap insurance would be to name the nine parameters in the new comment instead of describing them generically. That turns an invisible constraint into something a reviewer can check against a diff.

## A regression test is cheaper than the PR suggests

The description says a test "would mean carrying fixture directories" for a case CI doesn't otherwise cover. I don't think it has to. The schema entries are `format: path` with `exists: true`, so pointing a stanza's `salmon` attribute at any directory that already exists — `${projectDir}/assets`, say — satisfies validation without a real index, and under `-stub` the assertion is simply that `SALMON_INDEX` never appears in the trace. That covers the wiring this PR fixes, which is the part at risk of silent regression, even though it doesn't exercise functional index reuse.

Two caveats: the repo has no trace-based assertions today, so this would introduce a pattern, and it needs `--genome` plus a `genomes` block in the test config, which no existing test sets up.

## Loose end in the description

You report a pre-fix `Access to undefined parameter star_index` warning, but an exhaustive search shows `params.star_index` is read only in `main.nf` — the assignment at line 39 and the pass-through at line 72, where `PREPARE_GENOME` receives it as a workflow input rather than reading `params` itself. So the audit table's "star: read outside main.nf — no, correct" row checks out, and by rule 2 nothing should have warned for it. `bbsplit_index` is the one genuinely read in a stale-snapshot context, at `workflows/riboseq/main.nf:132`, so I suspect the warning was misattributed. Worth confirming: if there is a `star_index` reader I failed to find, that row of the audit needs revisiting.

## Minor

The CHANGELOG entry links `.../issues/195` where every other entry in `Fixed` links its PR, and where `4ac3345` on dev deliberately moved entries to PR numbers. I left it alone when resolving the conflict rather than fold a content change into a merge, so it's still outstanding.

For the record on that merge: `dev` had moved ahead by the #204 and #210 merges and both sides appended to the same point in the `Fixed` list, which was the only conflict. I kept all four entries in ascending order and pushed as a merge rather than a rebase, so your two commits are untouched. `dev` hadn't touched `main.nf` or `nextflow.config` since the merge base, and I checked the merged tree still has the assignment block above the `RIBOSEQ` and `PREPARE_GENOME` includes.

PR #209, opened 2026-07-27
fix: give Rp-Bp and PRICE the full reference plus novel genes under extended ORF analysis

### LINKED ISSUE #202: Hybrid GTF wiring inconsistency between Rpbp and PRICE

pinin4fjords:
Closes #202. Raised by @suhrig.

Rp-Bp and PRICE resolve overlapping ORFs themselves, so both need a multi-isoform annotation rather than the one-transcript-per-gene backbone. Under `--extended_orf_analysis` Rp-Bp got the hybrid GTF, which is built on that backbone, so the flag traded every isoform-specific ORF for the novel transcripts. PRICE got the full reference with no novel genes, so the flag did nothing for it.

Both now take `ch_full_hybrid_gtf`, the full reference with the novel intergenic genes appended, already built ahead of the caller dispatch for the ORF-level DTE denominator (#204) and catalogue gene-id resolution (#207). It is threaded into `ORF_CALLER_DISPATCH` as a fifth reference channel and shared by both callers, so nothing new is built.

Ribotricer and RiboCode are unchanged by design: Ribotricer uses the canonical backbone in both modes, and RiboCode's transcriptome BAM comes from the second STAR pass against the hybrid GTF, so its annotation has to match.

## Verified

Neither caller can be exercised by CI, which is why both are opt-in: Rp-Bp needs roughly 20-24h per replicate, and PRICE will not run on the minimal test data (#185), though it is fine at full scale. So the dispatch was run in isolation with three Ribo-seq samples, module stubs, and a distinct marker gene in each candidate annotation:

| Mode | `GEDI_INDEXGENOME` | `RPBP_PREPAREGENOME` | Rp-Bp per-sample stages |
| --- | --- | --- | --- |
| `--extended_orf_analysis` | full reference + novel | full reference + novel | 3/3 samples |
| default | full reference | full reference | 3/3 samples |

Neither stages the canonical backbone or the hybrid GTF. The 3/3 counts cover the #214 failure mode. This verifies the wiring, not the correctness of the calls.


FelixKrueger (commented):
Logic looks right. The thing I wanted to check was the moved `BUILD_FULL_HYBRID_GTF`: it's built under `extended_orf_active`, while the DTE consumers sit under `extended_orf_active && enabled_orf_callers && !params.skip_plastid`. Producer guard is strictly broader, so there's no path where `GFFREAD_FULL_HYBRID` or `STAR_GENOMEGENERATE_FULL_HYBRID` quietly get an empty channel. Sharing one build between the callers and the DTE denominator is a clear improvement over building it twice.

Two things before I approve, both about the base rather than the diff.

The conflict is in `orf_caller_dispatch` — your new `ch_full_hybrid_gtf` in the `take:` block lands on top of #214's value-channel contract, and the annotation-channel hunk overlaps as well. `workflows/riboseq/main.nf` should merge cleanly.

More importantly, the green run is from 12:18, and #214 (fix/200-channel-cardinality) merged at 14:01, with dev regenerating the extended-ORF snapshots at 12:42 in between. CI hasn't seen this combination. Since #214 is specifically about single-item queues serving only the first sample, and you're threading a new reference channel through that same plumbing, I'd rather see a re-run than reason about it — that failure mode is silent. Your `.first()` placement does look consistent with how #214 kept it on `ch_fasta_gtf` and dropped it on `ch_fasta_gtf_for_ribotish`, for what it's worth.

Merge dev in, re-run, and I'll approve.

Minor: the changelog entry points at `issues/202` rather than `pull/209`, and the `ext.prefix` change renames the PRICE index in default mode too (`price_index` → `price_index_reference`), which the entry doesn't mention.

pinin4fjords:
Merged dev in, re-run in flight.

Two notes. #207 landed the `BUILD_FULL_HYBRID_GTF` hoist on dev, so I'm no longer moving it and the diff is down to the routing alone.

And CI can't settle the #214 question for these two: they're opt-in *because* CI can't exercise them. Rp-Bp's 20-24h per replicate rules it out, and PRICE won't run on the minimal test data at all, `PriceOrfInference` crashes on chr20 and chr19+22 STAR output regardless of read depth (#185). PRICE is fine at full scale, the blocker is the test reference rather than the tool. Either way there's no pipeline test for the re-run to cover, so I ran the dispatch in isolation instead, three Ribo-seq samples, module stubs, a marker gene per candidate annotation:

| Mode | `GEDI_INDEXGENOME` | `RPBP_PREPAREGENOME` | Rp-Bp per-sample stages |
| --- | --- | --- | --- |
| `--extended_orf_analysis` | full reference + novel | full reference + novel | 3/3 |
| default | full reference | full reference | 3/3 |

All seven per-sample Rp-Bp stages run three times in both modes, so the new channel isn't a single-item queue. `FASTA_GTF_BAM_RPBP` also `.first()`s every reference it derives, so the contract holds at both levels.

Changelog points at `pull/209` now. On `ext.prefix` I dropped the change rather than documenting it: I'd justified the keying as keeping resume entries distinct, but the annotation differs between modes so the hash differs anyway. `price_index` stays `price_index`.

FelixKrueger (approved):
Conditions met — approving.

The resolution is better than mechanical: it keeps #214's value-channel contract on the original four and scopes an equivalent note to `ch_full_hybrid_gtf`. Both branches of `ch_fasta_gtf_multi_isoform` terminate in `.first()` after the combine, while `ch_fasta_gtf_for_ribotish` correctly has none — the distinction #214 drew is intact, which was my one real concern.

Re-run on 7a736247 is 27/27 with all 22 shard jobs actually executed, so the extended-ORF snapshots are covered against dev's regenerated baselines. Changelog entry now points at `pull/209` and dev's entries all survived.

Still not mentioned in the changelog entry: the `ext.prefix` change renames the PRICE index in default mode too (`price_index` → `price_index_reference`). Non-blocking — worth a clause if you're in the file anyway.

PR #210, opened 2026-07-27
Update trimgalore module to 2.3.0

FelixKrueger:
## Description of changes

Bumps the vendored `nf-core/trimgalore` module from **2.1.0 → 2.3.0**, regenerates the pipeline test snapshots for a behaviour change the new version carries, and corrects a stale description of Trim Galore! in the docs.

### Module bump

- `modules/nf-core/trimgalore/` updated via `nf-core modules update`, so the Seqera Wave container URIs (docker tag + Singularity blob) come from upstream rather than being hand-edited.
- `modules.json` pinned to `2a30bfb6a` — the upstream "Bump trimgalore to 2.3.0" commit (nf-core/modules#12331).
- All **six** vendored module files byte-match upstream at that SHA, so **no `.diff` patch file is required** and the module is not flagged as locally modified.

### ⚠️ Behaviour change: R2 poly-G trimming

Trim Galore 2.3.0 includes [TrimGalore#322](https://github.com/FelixKrueger/TrimGalore/pull/322) — *"fix(poly-g): trim R2's 3' poly-G tail (was incorrectly 5' poly-C)"*. **Poly-G trimming is on by default**, so this applies without any pipeline flag opting in.

| | Poly-G behaviour |
|---|---|
| 2.1.0 | poly-G tails from 3' of R1/SE, **poly-C heads from 5' of R2** ← incorrect |
| 2.3.0 | poly-G tails from 3' of **both R1 and R2** ← correct for 2-colour chemistry |

A controlled two-version run on identical input (500k read pairs, same args, threads pinned) confirmed the scope precisely:

- **Adapter and quality trimming are byte-identical** between the versions — same `Quality-trimmed` bases (63,073 / 73,548 bp), same `Reads with adapters` (177,340 / 178,393), same `Total written`.
- Only poly-G handling differs. R2 loses ~32 kb more bases, which moved pairs dropped by the 20 bp length cutoff from **6 to 2** on that sample.
- Consequence: per-sample read counts shift slightly for **paired-end** libraries, changing every downstream checksum. **Single-end libraries (including all Ribo-seq) are unaffected.**

### Snapshot regeneration

All 11 pipeline test snapshots were regenerated. The diff is **779 insertions / 778 deletions — pure md5 substitution, with no output added or removed**, and it lines up exactly with an R2-only fix:

| Category | Changed lines |
|---|---|
| RNA-seq, paired-end (SRX11780879–884) | 836 |
[...]

PR #211, opened 2026-07-27
fix(ribotish/predict): pin conda env to Python 3.13.9 to match container

### LINKED ISSUE #199: ribotish crash with -profile conda
### Description of the bug

When ribotish is installed through conda, a different Python version is used than what is used in the docker. It seems, the version is incompatible with ribotish. Appending the following line to `modules/nf-core/ribotish/predict/environment.yml` aligns the conda environment with the docker environment and fixes the problem:

```
  - python=3.13.9
```

```console
nextflow run riboseq/main.nf -profile conda,test --outdir=$PWD/results

  Process `NFCORE_RIBOSEQ:RIBOSEQ:ORF_CALLER_DISPATCH:RIBOTISH_PREDICT_INDIVIDUAL (SRX11780887)` terminated with an error exit status (1)
  NameError: name 'genomefapath' is not defined
```

### Relevant files

_No response_

### System information

pinin4fjords:
Closes #199. Thanks to @suhrig for the report, and for spotting that conda and the container resolve different Python versions.

## Mechanism

Ribo-TISH's `predict.py` stores per-run state (`genomefapath`, `tisbampaths`, `tisoffdict` and others) in module-level globals set inside `run()`. Its worker `_pred_gene`, mapped across a `multiprocessing.Pool` whenever `-p` is above 1, reads those globals back directly. That only works because the `fork` start method inherits the parent's already-mutated memory into each worker.

Python 3.14 changed the default start method on non-macOS POSIX from `fork` to `forkserver`. A forkserver worker is bootstrapped from a server process that re-imports the module fresh, so `run()` never executes there, the globals are never set, and referencing one raises exactly the reported `NameError`.

ribotish's bioconda recipe is `noarch: python`, so the same package runs under whatever Python the environment resolves. The container is fixed at 3.13.9, but `environment.yml` carried no pin, leaving conda free to drift past 3.14 independently.

**The pin is a ceiling holding Python below 3.14 in intent, not a preference for one patch release.** It is written as an exact version because nf-core's `environment-schema.json` rejects `<` and `>` in dependency specs, so `python<3.14` is not expressible. `3.13.9` matches the container exactly.

## This is an upstream Ribo-TISH bug

The pin is a workaround. The real defect is passing run-scoped state through module-level globals across a `Pool`, which is inherently unsafe under forkserver and spawn. I checked the Ribo-TISH tracker and found no existing issue for it; the closest is #3, a different and already-resolved Python 3.7 API problem. No upstream issue opened.

`ribotish/quality` has no `multiprocessing.Pool` usage at all, so it is not exposed and was left unpinned.

## Verified

- Reproduced the exact traceback: solved the real unpatched `bioconda::ribotish=0.2.8` environment, took the actual failing sample (`SRX11780887_chr20.bam`) and ran `ribotish predict -p 4` with the start method forced to `forkserver`. It fails at `predict.py:377` in `_pred_gene`, as reported.
- With the pin, the same environment resolves exact Python 3.13.9 and the identical invocation succeeds.
- Full `nextflow run . -profile conda,test` on this branch completes end to end: 357 tasks, 0 failed, 18m25s, with `RIBOTISH_PREDICT_INDIVIDUAL` (6/6) and `RIBOTISH_PREDICT_ALL` both passing.

Worth knowing: a bare `-profile conda,test` against `dev` does not currently reproduce the crash, because today's bioconda state happens to resolve 3.13.14 (pysam has no normal, non free-threaded 3.14 ABI build yet). That is channel-drift luck rather than design, and it is exactly the risk this pin removes.

`nf-core pipelines lint` was not run locally; CI covers it.

## Flagged for separate triage, not fixed here

`rpbp/*` has the identical unpinned `noarch: python` shape against a 3.13.13 container, and is a Python ORF caller in the same family, so it is worth checking for the same fork-reliant multiprocessing patterns. `ribocode/*` and `ribodetector` are also unpinned but sit on 3.10.x containers, so the urgency is lower.

The pin lives upstream in nf-core/modules#12438, and `ribotish/predict` is bumped here to that merged version rather than patched locally.


FelixKrueger (commented):
Thanks for the thorough write-up — the diagnosis and the evidence behind it made this easy to review. The change itself looks right to me, and I checked the parts that would have been expensive to take on trust. Two things block approval, both process rather than code.

## What I verified

- **nf-core/modules#12438 is genuinely merged**, and `a8e13ee` is its merge commit and an ancestor of `nf-core/modules` `master` — so the `modules.json` bump points at a released version rather than a branch tip.
- **The vendored module matches upstream exactly.** All five files (`environment.yml`, `main.nf`, `meta.yml`, `tests/main.nf.test`, `tests/main.nf.test.snap`) are byte-identical to `a8e13ee`, and the file set is identical too — so no local divergence and no missing patch file.
- **The pin is actually operative in this pipeline.** This was my main worry: if `RIBOTISH_PREDICT` were pinned to a pre-built conda lock file, editing `environment.yml` would silently change nothing under `-profile conda`. `conf/containers_conda_lock_files_amd64.config` turns out to cover only `FASTQC` and `MULTIQC`, so `RIBOTISH_PREDICT` does resolve from `environment.yml`. Fix confirmed effective.

Agreed too that the `transprofile` snapshot change is a separate, real fix rather than a test weakening — `imap_unordered` means the raw file md5 was never reproducible, and hashing sorted lines keeps content verification while dropping order sensitivity.

## Blocker 1 — merge conflict in `CHANGELOG.md`

The PR is currently `CONFLICTING`. At the merge base (`68002f9b`) the `Fixed` section ended at the `#194` entry; since then `dev` gained `#204`, renumbered `#206`→`#210`, and added a second `#210` line at that same spot, while this branch appends its `#199` line there too. Straight add/add conflict. `modules.json` should merge cleanly — `dev` only touched `trimgalore`, well away from `ribotish/predict`.

## Blocker 2 — CI hasn't validated the current head, and no nf-test has run at all

- Workflow runs exist only for `9495f48f` and `221c765e`. Nothing ran on `709b6b74` or `2e46777a`, which means **the commit that vendored the test changes and bumped `modules.json` has no CI coverage at all**.
- On `221c765e`, "Run nf-test" reports `success`, but its matrix job reports **`skipped`** — so zero nf-tests actually executed, while `confirm-pass` still went green.
- The `nf-core pipelines lint` comment in this thread is stamped for `221c765`, two commits stale.

I suspect the two blockers are cause and effect: `pull_request` workflows run against the ephemeral merge ref, and when the merge conflicts GitHub can't create that ref, so the triggers quietly don't fire. That would also explain why the `ci: retrigger checks` commit produced no run. A rebase should unblock CI on its own.

One thing worth watching on the re-run: a skipped matrix combined with a passing `confirm-pass` gate renders as all-green at the top level. If shard selection doesn't pick up `modules/nf-core/ribotish/predict/tests/`, this will look validated while having tested nothing — worth confirming the job actually ran rather than trusting the tick.

## Minor

The changelog entry links to `issues/199` rather than `pull/211`. All the other `Fixed` entries use `/pull/`, and `dev` carries 4ac3345e ("Point CHANGELOG entries at the actual PR number"), so this reads as a small inconsistency — though the `Added` section does use `/issues/` links, so it's clearly not a hard rule. Since resolving the conflict touches this line anyway, cheap to change if you agree.

No objection to the exact `python=3.13.9` pin given the schema constraint you describe, and agreed on leaving `ribotish/quality` alone and triaging `rpbp` separately.

## Before approving

Just a rebase on `dev` resolving the changelog conflict, and nf-test running green on the rebased head — with the ribotish/predict tests confirmed to have actually executed. The engineering here is sound; it only needs CI to confirm it against the code as it now stands.

FelixKrueger (approved):
Approving — and retracting one condition from my earlier review, which asked for something this repository cannot produce.

## Retracting the nf-test condition

I said I wanted "nf-test running green on the rebased head, with the ribotish/predict tests confirmed to have actually executed." That can't happen here, for two independent reasons:

1. `nf-test.config` deliberately excludes them — `ignore = ['modules/nf-core/**/tests/*', ...]`. The module's own tests are never run by this pipeline; testing is delegated to nf-core/modules.
2. `.github/workflows/nf-test.yml` runs `profile: [docker, singularity]`. There is no conda profile, so a conda-only `environment.yml` pin could not be exercised by pipeline-level tests either.

So `Found 0 related test(s)` → `total_shards: 0` → skipped matrix is the correct outcome for this change, not a failure. Sorry for setting a bar that couldn't be cleared — my earlier comment would have blocked this indefinitely.

The validation that matters did happen, upstream: nf-core/modules#12438 ran 64 checks including a full conda shard matrix (`x64 | conda | 10`–`16`) alongside docker and singularity, all green. That covers both the pin and the new sorted-lines transprofile hash.

## Verified on the current head (ab627e8f)

- Conflict resolved; `CHANGELOG.md` merged correctly, with dev's `#204` and both `#210` entries preserved and no stray merge markers.
- `nf-core` lint green — this time on the current head, not a stale commit.
- `pre-commit` green.
- All five files under `modules/nf-core/ribotish/predict/` are byte-identical to upstream `a8e13ee`, with an identical file set, and `modules.json` pins that exact SHA. The merge didn't disturb the vendored module.

Combined with your local `-profile conda,test` run, that's as much evidence as this change can be given. Approving.

## Two notes, neither blocking

The changelog entry still points at `issues/199` rather than `pull/211`. Entirely your call — I'm not going to hold the PR for it.

Separately, and not a criticism of this PR: `confirm-pass` reports success when the matrix is skipped, so a run that executes zero tests is indistinguishable from one where everything passed. Benign here, since the skip is legitimate, but it means a genuine shard-selection regression would also present as green. Probably worth raising against the shared workflow rather than any individual pipeline.

PR #212, opened 2026-07-27
fix(params): remove non-functional parameters and wire fastp_merge, save_merged_fastq and pseudo_aligner through

### LINKED ISSUE #197: Unused parameters

pinin4fjords:
Closes #197. Thanks to @suhrig for the report, which catalogued five parameters that either did nothing or were silently overridden.

## Breaking changes

Three parameters are removed:

| Removed | Why |
| --- | --- |
| `--min_mapped_reads` | Never read anywhere. A remnant of nf-core/rnaseq. |
| `--skip_pseudo_alignment` | Only decided whether a Salmon index was built. Setting it crashed strandedness inference. |
| `--skip_alignment` | Same category. An index-only run is not a supported mode, and every downstream stage depends on the STAR alignment. |

nf-schema only warns about unrecognised parameters, which is too quiet for flags whose whole purpose was to suppress work the pipeline does regardless. `validateInputParameters()` therefore errors at startup on all three, naming each one and what replaced it.

The `fail_mapped_samples` MultiQC section goes with `--min_mapped_reads`, since that filter was the only thing that ever populated it.

## `--pseudo_aligner` now works

It was read in exactly one place, to decide whether a Salmon index was needed, while the pseudo-alignment call site had `'salmon'` hardcoded. It now selects the tool used by `--te_quantification_method pseudo`, for both `salmon` (the default, unchanged) and `kallisto`.

- `PREPARE_GENOME` builds the matching TE index and emits it as `pseudo_index_te`, with the channel shape each quant module expects. `--kallisto_index` supplies a pre-built one.
- `--kallisto_quant_fraglen` and `--kallisto_quant_fraglen_sd` are new and required for the kallisto route, since kallisto cannot estimate a fragment length distribution from single-end reads and Ribo-seq libraries are single-end. The pipeline exits at startup rather than failing mid-run.
- `--extra_kallisto_quant_args` mirrors `--extra_salmon_quant_args`.
- TE pseudo outputs move to `quantification/<pseudo_aligner>_te_pseudo`, unchanged for the salmon default.

Deliberately not parameterised: the `'salmon'` at the `QUANTIFY_STAR_SALMON` call site, which runs with `alignment_mode = true` against the STAR transcriptome BAM and is not pseudo-alignment. Strandedness inference also always uses Salmon, so a Salmon index is still needed whichever pseudo-aligner is selected. Both are documented in `docs/usage.md`.

## `--fastp_merge` and `--save_merged_fastq` now take effect

Both were passed as hardcoded `false` into `FASTQ_QC_TRIM_FILTER_SETSTRANDEDNESS`.

`--save_merged_fastq` is worth calling out: it is a real parameter, declared in `nextflow.config`, present in the schema, honoured by the `CAT_FASTQ` `publishDir`, and documented in `docs/output.md` as producing `*.merged.fastq.gz`. The hardcoded `false` meant single-library samples bypassed `CAT_FASTQ` entirely, so that documented output could never be produced by any invocation.

## Kallisto modules

Both stay and `modules.json` is unchanged. `kallisto/quant` has `installed_by: ['quantify_pseudo_alignment']` and ships with the nf-core subworkflow the pipeline uses, so it cannot be dropped independently. `kallisto/index` is now used by `PREPARE_GENOME`.

## Overlap with #195

Both branches touch the `salmon_index_available` expression. Here it simplifies to `params.salmon_index as boolean`, since with `--skip_pseudo_alignment` gone the only way an index exists up front is for the user to have supplied one. Expect a conflict, to be resolved at merge time.

## Verified

- `nf-test tests/default.nf.test tests/te_pseudo_alignment.nf.test --profile test,docker`: both pass with no snapshot changes, confirming the renamed index channel and publish paths are behaviour-neutral for the salmon default.
- A full run with `--te_quantification_method pseudo --pseudo_aligner kallisto --kallisto_quant_fraglen 30 --kallisto_quant_fraglen_sd 5` completes, running `KALLISTO_INDEX_TE` and `KALLISTO_QUANT` and producing populated gene- and transcript-level TE matrices.
- `--pseudo_aligner kallisto` without the fragment length parameters exits at startup; each removed parameter triggers the startup error; a run with none of them is unaffected.
- A/B stub runs of `--trimmer fastp` with and without the two flags: with them, `FASTP` emits `*.merged.fastq.gz` and all 12 samples route through `CAT_FASTQ`; without, neither happens. Both arms stop at the same pre-existing stub limitation, so the difference is attributable to the flags.
- `nf-core pipelines lint`: 0 failures, 37 pre-existing warnings.

No CI coverage for the kallisto route: it needs fragment length parameters no existing test profile sets, so it would need a new test file. Verified manually instead.


FelixKrueger (commented):
Careful piece of work. I went looking for the usual ways a breaking parameter removal goes wrong and didn't find them: I grepped the whole tree at your head and the three removed names survive only in the `removedParamsError()` map — nothing left in `conf/test_full.config`, which PR CI never runs. `containsKey()` is the right mechanism there.

The `QUANTIFY_PSEUDO_TE` wiring also checks out, which was my main worry given the kallisto path has no CI cover: 13 arguments matching the nf-core subworkflow's `take:` block in order, both fraglen values in the right slots. And `params.pseudo_aligner` is never used as a boolean anywhere, so flipping the default from `null` to `salmon` is safe — the interpolated publish paths resolve to the old `salmon_te_pseudo`, which the green snapshots confirm.

One thing to decide before this merges. #208 landed at 15:01, 66 minutes after your CI ran at 13:55, and it removed `salmon_index = null` from `nextflow.config` in favour of `params.salmon_index = getGenomeAttribute('salmon')` in `main.nf`. You're adding `kallisto_index = null` to that same block, and `params.kallisto_index` to the same region of `main.nf`. So: should `kallisto_index` follow the genome-stanza pattern too? Someone who can put `salmon` in a stanza will expect `kallisto` to work the same way, and it's much cheaper to settle now than after the parameter ships. Your call, but the merge forces the question either way. dev also gained #211 and #214 in that window, so the re-run isn't optional.

Worth adding while you're there: nothing validates `pseudo_aligner_kmer_size` for the kallisto route. kallisto wants an odd k no greater than 31, the docs say so, but `--pseudo_aligner kallisto --pseudo_aligner_kmer_size 24` still dies mid-run at `KALLISTO_INDEX_TE`. That's the same failure mode `kallistoPrerequisitesError()` exists to prevent, and the schema could carry `multipleOf`/`maximum` for it.

Minor: the changelog entry points at `issues/197` rather than `pull/212`. Unrelated to you, but `if ('salmon' in prepare_tool_indices)` in `prepare_genome` is unreachable — nothing ever appends `'salmon'` to that list, on dev either — and it probably wants reconciling with what #208 just changed.

pinin4fjords:
Thanks for the review. Merged dev in and addressed the two real points:

- `kallisto_index` now follows the same `getGenomeAttribute('kallisto')` pattern as `salmon_index`/`sortmerna_index` from #208, for consistency.
- `kallistoPrerequisitesError()` now rejects an even or >31 `--pseudo_aligner_kmer_size` before building the kallisto index, matching what the docs already said.

Leaving the changelog `issues/` vs `pull/` link style alone, it matches plenty of existing entries.

FelixKrueger (commented):
Re-checked at `35b81b8` rather than taking the reply on trust. Both points are properly addressed, and a third I'd flagged is fixed too.

**Genome-stanza pattern for `kallisto_index`** — correct against both mechanisms #208 turns on:

- `kallisto_index` is absent from `nextflow.config`, so nothing shadows the assignment.
- `params.kallisto_index = getGenomeAttribute('kallisto')` sits at `main.nf:42`, above `include { RIBOSEQ }` at 45, so every consumer inherits the resolved value rather than a pre-assignment snapshot.
- The schema entry carries no default and `exists: true`, matching the other nine, so there's no config/schema mismatch for lint to find.

`format: file-path` rather than salmon's `path` is right, not an inconsistency — a kallisto index is a single `.idx`, which is what `prepare_genome`'s `take:` block documents, while salmon's is a directory.

**Kmer validation** — scoped correctly. It early-returns unless `--pseudo_aligner kallisto` and `--te_quantification_method pseudo`, and it's gated on `!params.kallisto_index` so it only fires when an index will actually be built. K-mer size is meaningless with a pre-built index, and a guard that ignored that would reject valid runs.

**`prepare_tool_indices`** — the unreachable `'salmon'` branch is gone, the list now has exactly matching consumers, and salmon/kallisto index building moved onto the explicit `build_salmon_index_for_strandedness` and `build_te_pseudo_index` booleans. Cleaner than what I asked for.

## One thing before merge

The stanza table in `docs/usage.md` is a row short now. #208 added it listing the recognised attributes; `main.nf` resolves ten and the table lists nine, so `kallisto = '/path/to/index.idx'` in a stanza is honoured but undocumented:

| `kallisto`         | `--kallisto_index`   |

Which points at the wider problem. That set of parameters is a load-bearing invariant — every one of them must stay absent from `nextflow.config` or its stanza attribute silently stops working — and nothing tracks it: no test, no lint rule, and CI never exercises the stanza path because `conf/test.config` sets `fasta`/`gtf` directly and no test passes `--genome`. It went stale the first time someone extended it, within a day of #208 landing. Naming the parameters in that `nextflow.config` comment instead of describing them generically would at least put the constraint where the next person editing the block will see it.

## Optional

The kmer guard has no lower bound. `pseudo_aligner_kmer_size` carries only `type: integer` and `default: 23` — no `minimum` — so `--pseudo_aligner_kmer_size -3` passes schema validation *and* passes the new guard, because in Groovy `-3 % 2` is `-1` rather than `0`, and `-3 > 31` is false. It then dies inside `KALLISTO_INDEX_TE`, which is the failure mode the guard exists to prevent. A schema `"minimum": 1` closes it declaratively and produces a sensible message for free; widening the condition to `< 1` would do as well.

Changelog link style: agreed, dropping it.

Not approving yet purely for the table row. Worth having CI green before merge in any case, given the kallisto path has no test cover and the dev merge brought in #211 and #214.

FelixKrueger (approved):
Both points are closed, and the second one better than I asked for.

The stanza table gained the `kallisto` row, and the `nextflow.config` comment now enumerates all ten parameters rather than describing them generically. I cross-checked the three lists mechanically — the parameters assigned in `main.nf`, the rows in the `docs/usage.md` stanza table, and the names in that comment are now identical sets. That turns the invariant into something a reviewer can verify by eye against the diff, which is what was missing when the table went stale in the first place.

Putting the bound in the schema rather than widening the Groovy condition is the better of the two options. nf-schema rejects the value during `validateParameters()`, before any process is submitted, and it covers the salmon route too — that parameter is shared, and the salmon path never reaches `kallistoPrerequisitesError()` at all. Recording the `-3 % 2 == -1` reasoning in the commit message is the right place for it, so nobody later concludes an odd-check alone would have been enough.

Carrying forward from the previous round, verified at this head rather than assumed: `kallisto_index` stays absent from `nextflow.config`, `params.kallisto_index = getGenomeAttribute('kallisto')` sits above the `RIBOSEQ` include, the schema entry carries `exists: true` with no default, `format: file-path` correctly reflects that a kallisto index is a single `.idx` unlike salmon's directory, and the unreachable `'salmon' in prepare_tool_indices` branch is gone with index building moved onto the explicit booleans.

One thing for the record rather than for action: nothing in the test suite sets `--pseudo_aligner kallisto`, so green CI here does not exercise the `QUANTIFY_PSEUDO_TE` wiring, the fraglen requirement, or the new k-mer guard. Those have been checked by reading the code against the subworkflow's `take:` block and by reasoning about the parameter combinations — which is a real check, but not the same as a run. Worth knowing if the kallisto route later misbehaves: the argument order and the guard conditions are the parts no test would have caught.

PR #213, opened 2026-07-27
Anchor subworkflow withName selectors so the gffcompare class-code filter actually runs

### LINKED ISSUE #201: output file `stringtie_merge/stringtie_merge.gtf` empty
### Description of the bug

The file `stringtie/stringtie_merge.gtf` is always empty. The generating process `GTF_HYBRIDMERGE_GFFCOMPARE:GAWK_FILTER` is configured in two places:

1. Using a full-name selector here:
https://github.com/nf-core/riboseq/blob/68002f9b8da43389a23d8f9b4f6ac89dc9d95409/subworkflows/nf-core/gtf_hybridmerge_gffcompare/nextflow.config#L23
2. Using a glob selector here:
https://github.com/nf-core/riboseq/blob/68002f9b8da43389a23d8f9b4f6ac89dc9d95409/conf/modules.config#L1246

The first configuration contains a critical setting for the gawk filter: it selects novel transcripts by means of passing `-v codes_csv='u'`. This config is lost when the second config is present in the form of a glob selector. I'm not exactly sure why. Maybe it has to do with the process being called from within a subworkflow. When I change the second config to a full-name selector, then the two configs are merged in an additive manner as one would expect intuitively.

Two other processes (`GTF_HYBRIDMERGE_GFFCOMPARE:BEDTOOLS_INTERSECT` and `GTF_HYBRIDMERGE_GFFCOMPARE:GAWK_CONCAT` also seem affected by this, although the effects may be milder here, since only the output file naming is affected.

```console

```

### Relevant files

_No response_

### System information

pinin4fjords:
Closes #201. Reported by @suhrig.

## Root cause

Nextflow matches `withName` as a full-string regex against the fully qualified process name. Two subworkflow-shipped configs used selectors of the form `SUBWORKFLOW:PROCESS` with no leading `.*`, so they matched only under the subworkflows' own nf-tests and never inside the pipeline, where the qualified name is `NFCORE_RIBOSEQ:RIBOSEQ:...:SUBWORKFLOW:PROCESS`. Confirmed with a three-level probe pipeline on Nextflow 26.04.6.

This is not an interaction with `conf/modules.config`. The two selectors are distinct keys and merge additively, but only the `.*:`-prefixed one ever matched.

## Impact

`gtf_hybridmerge_gffcompare` is the severe one. `GAWK_FILTER` lost `-v codes_csv=...`, so the class-code filter built an empty keep-set and emitted nothing. The hybrid GTF was the canonical backbone with zero novel transcripts, and everything gated on `--extended_orf_analysis` (Ribo-TISH predict, Ribotricer, RiboCode, the hybrid transcriptome and the second STAR pass) ran against it. Extended ORF analysis could never call an ORF on a novel transcript.

Evidence: before this change `hybrid_reference.gtf` had md5 `fb1c85fe...` in all three tests that build it, despite three different novel-transcript inputs. After, all three differ.

The empty `stringtie/stringtie_merge.gtf` in the report is a second, separate symptom. `GAWK_FILTER` also lost `ext.prefix` and so published to the same path `STRINGTIE_MERGE` uses, overwriting it. The merged assembly itself was always produced and always reached gffcompare over the channel.

`orftable_fasta_gtf_buildorfcatalogue` was cosmetic: `conf/modules.config` already duplicated its `ext.args`, and the lost `ext.prefix` values name unpublished work-directory intermediates.

## On #204

#204 notes that StringTie finds no novel transcripts in the chr20 data. That was itself a symptom of this bug: StringTie does find 44 class-`u` transcripts, and they were discarded before reaching the hybrid GTF. #204's conclusion still holds for a different reason, since none of those loci carries enough Ribo-seq signal to call an ORF. The cohort catalogue and per-ORF P-site counts are byte-identical before and after, and no prediction references an `MSTRG.*` transcript.

## Changes

- Anchor the selectors in both subworkflow configs with `.*` so they apply standalone and in-pipeline.
- Drop the `ext.args` duplicated into `conf/modules.config` for the three orf-catalogue processes, leaving the subworkflow config as the single source. The strand-aware `BEDTOOLS_INTERSECT` override stays, being a deliberate pipeline-level choice.
- Pin `ext.prefix = 'hybrid_reference'` for `GAWK_CONCAT` so the hybrid GTF keeps its documented published name.
- `docs/output.md`: correct the filter and blacklist filenames, and the claim that the hybrid GTF is coordinate-sorted.

Both subworkflows carry the fix upstream (nf-core/modules#12435 and #12434), since every consumer had the same bug, and are bumped here to those merged versions rather than patched locally.

## Tests

`stringtie/stringtie_merge.gtf` is now the real assembly, and `stringtie/<source>_filtered.gtf` is a new published file.

StringTie assigns `MSTRG.*` ids in a run-varying order, so the merged assembly and its derivatives are content-unstable. That was hidden while the filter output was empty and the hybrid was pure canonical. Those four artefacts move to `tests/.nftignore_orf`, replaced by direct assertions that the hybrid GTF holds at least one `class_code "u"` transcript and nothing of any other class, which is a stronger guard against this regression than an md5. Ribo-TISH outputs on the StringTie path shift because predict now sees 44 extra transcripts; both StringTie-path tests agree on the new values.

## Verified

All three extended-ORF tests pass on Nextflow 26.04.6. Filter output inspected directly: 44 transcripts on the StringTie path and 1 on the `--novel_gtf` path, every one `class_code "u"`. Determinism checked across two independent runs, where 264 of 274 shared md5s match and the 10 that do not are the ones now ignored. `nf-core pipelines lint` reports 0 failures.

Not verified: the `--rrna_blacklist` path, since no test supplies a blacklist, and whether novel-transcript ORFs are recovered at real data scale.


FelixKrueger:
Diagnosis checks out, but two things are blocking.

**Verified:**

- Pre-fix snapshots recorded the filtered GTF at `d41d8cd98f00b204e9800998ecf8427e` (empty file), so the class-code filter really was dropping everything — `codes_csv` never reached `filter.awk`.
- `GAWK_FILTER` was additionally publishing to `stringtie/stringtie_merge.gtf`, same path and filename as `STRINGTIE_MERGE` (both get `[id: 'stringtie_merge']`), and the empty file was the one that landed. Now distinct as `stringtie_merge_filtered.gtf`.
- `ext.prefix = 'hybrid_reference'` is load-bearing, not cosmetic: once the subworkflow config starts applying, `${meta.id}_hybrid` would rename the output and break the documented path plus the `hybrid_gtf` emit.
- Both vendored subworkflows are byte-identical to the pinned SHAs (`5ca7e22`, `23ca2b3`), so the modules-up-to-date check will pass.
- Anchoring sweep is complete — the other three subworkflow configs included by `nextflow.config` were already anchored, and the remaining unanchored selectors are all under `tests/`.

**Blocking:**

1. Conflicts with `dev` in `CHANGELOG.md` and all three ORF snapshots. Since #204 and #210 landed, `dev` moved those snapshots by roughly +196/-60 lines each while this branch moved them by +13/-21. The Trim Galore 2.3.0 bump changes md5s downstream and #204 rewrote the same ORF-level DTE outputs, so these need re-recording on top of `dev` rather than a manual conflict resolve.
2. No CI on `4ca6f11`. Last green run was `6a0d4de`; the lint comment above is for `80922fc`, two commits back. `4ca6f11` is the commit that bumps `modules.json`, so the up-to-date check is precisely what hasn't run.

**Nits:**

- `dotseq.nf.test` only got the `count > 0` assertion. The strict `every { !line.contains('class_code') || line.contains('class_code "u"') }` check applies there too — default `gffcompare_class_codes = 'u'` and the test doesn't override it.
- `hybrid_reference.gtf` lost its md5 in `novel_gtf.nf.test`, where the input GTF is static and the output should be reproducible. Shared `.nftignore_orf` makes that awkward to scope per-test, so fine if deliberate.
- CHANGELOG entry points at `issues/201`; convention (and `4ac3345` on `dev`) is the PR number.

suhrig:
Used the opportunity to test the `--rrna_blacklist` path, which I had never tested before: It works. Will leave @FelixKrueger the honours of formally approving this once he is satisfied, too.

FelixKrueger (approved):
Both things I blocked on are sorted, so this is good to go from my side.

The conflicts are resolved by merging `dev` in rather than hand-editing, so the ORF snapshots now sit on top of #204, #210, #217 and #220 — and green CI on this head is what actually proves they were re-recorded rather than patched together.

The modules check has run too. I'd checked the same thing by hand first: all six vendored files across both subworkflows are byte-identical to the pinned SHAs (`5ca7e226`, `23ca2b32`), with no patch entries, so nothing was modified locally.

Nits are dealt with — the strict class-code assertion is in `dotseq.nf.test` now, and the CHANGELOG points at the PR. Leaving `hybrid_reference.gtf` without an md5 in `novel_gtf.nf.test` is fine by me: `.nftignore_orf` is shared across all three extended-ORF tests, and the class-code assertions guard the actual regression better than an md5 did.

One thing for the record rather than a request. Between this PR and the two it unblocks, the novel-transcript path now has no snapshot coverage at all — `stringtie_merge.gtf`, `stringtie_merge_filtered.gtf`, `hybrid_reference.gtf`, `salmon_hybrid/tx2gene.tsv` and the ribotish transprofiles are ignored here, and `dte/orf_level/orf_combined_counts.tsv` already was. Ignoring them is the right call given you showed the MSTRG ids move between runs; re-recording a genuinely non-deterministic file would just defer the failure.

But it leaves a gap worth naming. #204 and #207 both built machinery for ORFs on novel transcripts, and this PR shows the hybrid GTF held zero novel transcripts the whole time, so neither was ever exercised. Now 44 class-`u` transcripts get through and, as you note, no prediction still references an `MSTRG.*` id. So that feature ships on your real-data checks rather than on CI. Nothing to fix here, but if the chr20 fixture will never call an ORF on a novel transcript, a fixture that does would be worth having at some point.

Thanks for going after the root cause on this one. The three-level probe pipeline and the before/after `hybrid_reference.gtf` md5s made the diagnosis straightforward to check.

PR #217, opened 2026-07-28
Fix GEDI genome index not being self-contained (breaks PRICE on cloud executors)

FelixKrueger:
## Description

Re-pins `gedi/indexgenome` and `gedi/price` to [nf-core/modules#12451](https://github.com/nf-core/modules/pull/12451), which makes the GEDI genome index relocatable so `--run_price` works on cloud executors.

`gedi/indexgenome` wrote an index that only resolved in the directory it was built in. GEDI records the `-s` path in the `.fi` it writes: a bare relative name when the FASTA sits beside the `.fi`, an absolute path otherwise. Passing a staged input at the task work directory root therefore recorded an absolute path to a file outside the emitted index, which Fusion never materialises, and PRICE died in `PriceCodonInference`:

```
Could not read sequence from file /fusion/s3/.../price_index/<genome>.fi!
/fusion/s3/.../<genome>.fa (No such file or directory)
```

Upstream fixes both halves. `indexgenome` indexes from a copy inside the index directory, so GEDI records a relative name that travels with the index. `gedi/price` rewrites the `.oml`'s absolute member paths onto the staged index, into the task directory rather than through the staged directory, since writing through a staged directory lands back at the upstream task's storage prefix under Fusion. Together the index survives being copied away from the work directory it was built in.

Nothing on the pipeline side needs to change: no new inputs, and `conf/modules.config` is untouched.

## Temporary CI change

`.github/workflows/nf-test.yml` points nf-test at GitHub-hosted runners, because the nf-core self-hosted fleet is currently down and every riboseq nf-test run was queueing indefinitely without ever being assigned a runner. Same workaround as [nf-core/sarek](https://github.com/nf-core/sarek) took. Two consequences worth knowing:

- This should be reverted once the self-hosted fleet is back.
- The `volume=120g` disk request is dropped in the process. The full suite passed on hosted runners (14 shards x 2 Nextflow versions), so hosted disk is sufficient today, but that margin is not guaranteed.

Other riboseq branches are blocked by the same outage, so this swap may be worth landing on `dev` rather than carrying it here.

## Things worth knowing

Upstream also swapped `gedi/price`'s pinned `orfs.tsv` row count for the header columns plus a row-count band, because the count varies with container engine and thread count. That test change comes along with the re-pin.

`--run_price` is off by default and no pipeline test enables it, so nothing in CI covers this path, and a local run works either way because work directories share a filesystem. Verified by running PRICE against a copied-out index whose indexing work directory had been deleted.

## PR checklist

- [x] Description of changes, with reason
- [x] `CHANGELOG.md` is updated
- [x] `nf-core modules lint gedi/indexgenome gedi/price` passes (0 failures; warnings unchanged from upstream)
- [ ] Tests: none added here, the fix and its tests live in nf-core/modules


modules/nf-core/gedi/indexgenome/main.nf
```diff
@@ -23,8 +23,18 @@ process GEDI_INDEXGENOME {
...
     """
     mkdir -p ${prefix}
+
+    # GEDI records absolute paths inside the .oml and .fi it writes, including one
+    # pointing at the FASTA given to -s. A staged input sits at the task work
+    # directory root, which is not part of the emitted index directory, so on
+    # executors where work directories are not a shared filesystem (AWS Batch with
+    # Fusion, and cloud executors generally) that path is never materialised and
+    # downstream PRICE dies in PriceCodonInference reading sequence. Index from a
+    # copy inside the index directory so the recorded path travels with the index.
```
pinin4fjords: Most of this is PR description rather than code comment. The part a reader can't infer is GEDI's path behaviour:

```suggestion
    # GEDI records the -s path in the .fi it writes: a bare relative name when the FASTA
    # is co-located with the .fi, an absolute path otherwise. Index from a copy inside the
    # index directory so the reference resolves wherever the index is later read.
```

CHANGELOG.md
```diff
@@ -41,6 +41,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - [#210](https://github.com/nf-core/riboseq/pull/210) - Correct the `--cores` note in the usage and output docs, which dated from the 0.6.x Perl era: worker cores are capped at 8 (not 4), multi-core trimming needs more than 4 CPUs for single-end and more than 5 for paired-end data, and the 2.x N+4 thread model is described. Also fixes a 404 changelog link (`Changelog.md` → `CHANGELOG.md`) ([@FelixKrueger](https://github.com/FelixKrueger))
+- [#217](https://github.com/nf-core/riboseq/pull/217) - Patch `gedi/indexgenome` so the GEDI genome index is self-contained. GEDI records absolute paths in the `.oml`/`.fi` it writes, including one to the source FASTA, which was staged at the task work directory root and so never became part of the emitted index. On executors where work directories are not a shared filesystem (AWS Batch with Fusion, and cloud executors generally) that path was never materialised, and `--run_price` failed in `PriceCodonInference` with `Could not read sequence from file`. The FASTA is now copied into the index directory and indexed from there, so the recorded path travels with the index ([@FelixKrueger](https://github.com/FelixKrueger))
```
pinin4fjords: One sentence is enough, with the executor claim scoped to Fusion:

```suggestion
- [#217](https://github.com/nf-core/riboseq/pull/217) - Patch `gedi/indexgenome` to index from a copy of the FASTA placed inside the index directory, fixing `--run_price` failing in PRICE's `PriceCodonInference` on AWS Batch with Fusion, where the path GEDI recorded for the staged input was never materialised ([@FelixKrueger](https://github.com/FelixKrueger))
```

pinin4fjords (changes requested):
Verified: reproduced the failure locally (unpatched with the work-root FASTA removed fails in `PriceCodonInference` with your exact error, patched passes with identical ORF calls). The mechanism is co-location: GEDI only writes a bare relative FASTA path into the `.fi` when the two share a directory, so the copy is necessary rather than one option among several.

Two asks. It fixes Fusion, not cloud executors generally, since a published index with the indexing work dir gone still fails at `.oml` load. And it should land in nf-core/modules rather than as a pipeline patch, together with a one-line `.oml` rewrite in `gedi/price` that makes the index executor-agnostic, after which we re-pin here and drop the patch file. Happy to do the upstreaming if you'd rather not.

pinin4fjords:
Concretely, the `gedi/price` half is:

```bash
sed "s|file=\"[^\"]*/|file=\"$PWD/${index}/|g" ${oml} > local_reference.oml
```

then `-genomic local_reference.oml`. Writing the rewrite into the task directory rather than the staged index keeps Fusion from writing back upstream. I ran it against a relocated index with the indexing work dir deleted and PRICE completed, so your FASTA copy plus this makes the index portable. No new input needed on `gedi/price`, since the copy already put the FASTA inside the index.

pinin4fjords:
I'll draft an nf-core/modules PR

pinin4fjords (approved):
Both asks are in. The fix is now upstream as nf-core/modules#12451, carrying your FASTA copy plus an `.oml` path rewrite in `gedi/price`, so this PR is a straight re-pin of `gedi/indexgenome` and `gedi/price` with the patch file dropped. The index is relocatable rather than only resolvable under Fusion, verified against a copied-out index whose indexing work directory had been deleted.

`dev` is merged in and CI is green. Changelog entry condensed to one sentence.

PR #220, opened 2026-07-28
Strip meta from BAM_DEDUP_UMI multiqc files (--with_umi runs fail at MultiQC)

FelixKrueger:
## Description

Runs with `--with_umi` fail at MultiQC:

```
Not a valid path value type: java.util.LinkedHashMap ([id:CD70plus_D4_3_ribo, strandedness:unstranded, sample_type:riboseq, ...])
```

`BAM_DEDUP_UMI` emits `multiqc_files` as `[ meta, file ]` tuples. It's built by mixing `ch_genomic_dedup_log` with the genome stats, flagstat and idxstats channels, and all of those carry a meta. The emit comment says `channel: file`, which isn't accurate.

`MULTIQC`'s input does `ch_multiqc_files.flatten().collect()`, so `flatten()` splits each tuple and the meta map ends up being passed where a path is expected.

Fix maps the meta away at the mix, which is what line 202 already does for `FASTQ_QC_TRIM_FILTER_SETSTRANDEDNESS.out.multiqc_files`.

I fixed it on the pipeline side rather than in `bam_dedup_umi`, since that subworkflow is shared with other pipelines and changing its emission shape could break them. The misleading emit comment would be worth correcting upstream separately.

## Coverage

Nothing in CI exercises this. `with_umi` defaults to false and no test enables it, which is why it went unnoticed. Adding coverage would mean turning UMI handling on somewhere in the test matrix, which seemed beyond the scope of the fix, but I'm happy to do it if you'd prefer.

No overlap with #214: that touches this file only at lines 242, 473 and 609.

## PR checklist

- [x] Description of changes, with reason
- [x] `CHANGELOG.md` is updated
- [x] `nextflow lint` is clean on the changed file (the one error it reports is pre-existing in `utils_nextflow_pipeline/main.nf` and also present on dev)
- [ ] Tests: none added, see the coverage note above


workflows/riboseq/main.nf
```diff
@@ -279,8 +279,11 @@ workflow RIBOSEQ {
         ch_transcriptome_bam = BAM_DEDUP_UMI.out.transcriptome_bam
         ch_genome_bam_index  = BAM_DEDUP_UMI.out.index
 
+        // BAM_DEDUP_UMI emits multiqc_files as [ meta, file ] despite its
+        // 'channel: file' comment, so drop the meta here. MULTIQC flattens this
+        // channel, which would otherwise pass the meta map itself as a file.
```
pinin4fjords:
```suggestion
```

This is done a lot above, we don't need to re-litigate the reasoning

pinin4fjords (approved):

PR #222, opened 2026-07-31
Build the ORF catalogue from --extended_orf_analysis alone

FelixKrueger:
`--skip_stringtie` defaults to `true`, so `--extended_orf_analysis true` on its own built no ORF catalogue, no `orf_quantification/` and no ORF-level DTE — silently. One predicate was gating both "route hybrid annotation into the callers" (which does need novel transcripts) and "build the catalogue" (which only needs callers and a GTF). This splits them, so the catalogue now follows `--extended_orf_analysis` alone. Caller routing is unchanged. No change to default runs, and existing snapshots are untouched.

<details>
<summary>Implementation notes and rationale (AI-assisted — keep or bin as you see fit)</summary>

### What changed

`extended_orf_active` keeps gating caller routing. A new `orf_catalogue_active` (`extended_orf_analysis && enabled_orf_callers`) gates the catalogue, ORF quantification and ORF-level DTE. `BUILD_FULL_HYBRID_GTF` now runs when either is active, so the catalogue always has an annotation.

`enabled_orf_callers` is hoisted above the hybrid-GTF guard, since that guard now consumes it.

### Why build the GTF unconditionally rather than fall back to `ch_gtf`

Two reasons, both found in review:

1. `main.nf` reads `BUILD_FULL_HYBRID_GTF.out` **directly** in the ORF-DTE block, not via `ch_full_hybrid_gtf`. Leaving the process gated while opening that block fails at workflow construction with `Access to 'BUILD_FULL_HYBRID_GTF.out' is undefined`. A grep for the channel name doesn't find those references.
2. A plain `ch_gtf` fallback loses the "catalogue annotation ⊇ every caller's annotation" guarantee when `--canonical_gtf` is user-supplied — the schema recommends MANE Select for human, whose IDs don't exist in an Ensembl GTF. `orfnormalise` then degrades silently: multi-exon ORFs collapse to single intron-spanning blocks, which feeds wrong sequence into `BEDTOOLS_GETFASTA -split` and produces wrong peptides.

With no novel source, `ch_hybrid_gtf` is the canonical backbone and the awk emits the reference plus any canonical-only transcripts, so the superset property holds by construction. `BUILD_FULL_HYBRID_GTF` publishes nothing without `--save_reference`, so running it more often has no snapshot impact.

### RNA denominator

Without a novel source there are no novel genes, so the ORF-level DTE denominator reuses the primary Salmon matrix rather than building a second STAR index and requantifying every RNA-seq library. `dte_counts_prep` already handles being given a matrix carrying Ribo-seq columns. Side benefit: the ORF-level denominator becomes exactly the gene-level one.

### Incidental fixes

- `dotseqPrerequisitesError` gated on `extended_orf_active`, so ORF-level dotseq would have errored on the newly valid combination — and deleting only the message clause would have produced `requires: .` with an empty list. The gate itself now uses the new predicate.
- The `--skip_plastid` warning required `novel_source_configured`, so it was suppressed in exactly the configuration that now needs it.
- New warning when `--orf_min_callers` exceeds the enabled caller count; the schema recommends `2`, and a single-caller run otherwise publishes a header-only consensus view with no indication. Note `-profile test` is single-caller (`skip_ribocode = true`).
- Removed `orf_agreement_min_callers`, `rank_aggregation_callers`, `ch_enabled_orf_callers` and `ch_rank_aggregation_callers` — no consumers anywhere; the live threshold is `params.orf_min_callers`.

### Docs

`nextflow_schema.json` and `docs/usage.md` both stated the flag was a no-op without a novel source; that is now true only of caller routing. `docs/output.md` already described the post-change behaviour, so this is partly the code catching up with the docs.

Added a note that catalogue `orf_id` values are assigned by coordinate order, so adding novel transcripts renumbers everything downstream — runs are comparable on coordinates, never on `orf_id`.

### Verification

New test `tests/orf_catalogue_no_novel_source.nf.test` covers the configuration nothing covered before: `extended_orf_analysis = true`, `skip_stringtie = true`, no `novel_gtf`. It asserts the catalogue and `orf_quantification/` are non-empty, and that `hybrid_star/` and `quantification/salmon_hybrid/` are absent. Its snapshot needs recording on CI.

Checked under `-preview`, which exercises workflow construction:

| Configuration | Result |
|---|---|
| `extended_orf_analysis` + `skip_stringtie`, no `novel_gtf` | constructs; warning reworded |
| novel source (`skip_stringtie false`) | unchanged |
| default (no extended) | unchanged |
| `dotseq` + extended + `skip_stringtie` | proceeds (previously errored) |
| extended + `skip_plastid` | plastid warning fires (previously suppressed) |
| `--orf_min_callers 2`, one caller | new warning fires; silent at the default of 1 |

`nextflow lint` reports no new errors. Not yet run end to end — the full test suite is the remaining check.

</details>


subworkflows/local/utils_nfcore_riboseq_pipeline/main.nf
```diff
@@ -191,16 +191,23 @@ def validateInputParameters() {
     dotseqPrerequisitesError()
     kallistoPrerequisitesError()
 
-    // --extended_orf_analysis is a no-op without a novel-transcript source
-    // (StringTie or a user-supplied --novel_gtf); warn rather than error so
-    // flags can be composed incrementally.
+    // Without a novel source only caller routing is a no-op; the catalogue still builds.
```
pinin4fjords: **Severity: low**: This comment uses change-relative wording. State the invariant directly so it remains understandable without the change history.

```suggestion
    // Without a novel source, caller routing is a no-op; catalogue construction is independent.
```

tests/orf_catalogue_no_novel_source.nf.test
```diff
@@ -0,0 +1,62 @@
...
+
+        when {
+            params {
+                outdir                = "$outputDir"
+                extended_orf_analysis = true
+                skip_stringtie        = true
+                // novel_gtf deliberately unset: this is the configuration that
+                // previously produced no catalogue at all. -profile test supplies
+                // contrasts and leaves skip_plastid false, which is what exercises
+                // the ORF quantification and ORF-level DTE paths.
```
pinin4fjords: **Severity: low**: This comment narrates earlier behavior. Keep only the test configuration and the behavior it exercises.

```suggestion
                // novel_gtf deliberately unset to exercise catalogue construction
                // without a novel-transcript source. -profile test supplies contrasts
                // and leaves skip_plastid false, exercising ORF quantification and
                // ORF-level DTE.
```

pinin4fjords (commented):
AI-assisted review (Codex, on behalf of the requester). No blocking correctness or Nextflow channel-semantics issues found. These two low-priority suggestions make the new comments independent of implementation history. Approval should wait for the new snapshot and the remaining nf-test jobs.

FelixKrueger:
The ❌ nf-test comment above is stale. It refers to [run 30663366702](https://github.com/nf-core/riboseq/actions/runs/30663366702), which was cancelled mid-flight when the snapshot commit was pushed. That shard failed only because `tests/orf_catalogue_no_novel_source.nf.test` had no `.snap` yet.

Current head `2d2e28c` is green — all 34 checks pass, and the new test passes on both Nextflow legs: `25.10.4` in 2515s and `latest-everything` in 2527s.

Worth flagging that the sticky comment cannot clear itself: `.github/workflows/nf-test.yml` only writes `pr-comment/comment.md` when failure fragments exist, so a run that goes green produces nothing for the poster to publish and the previous failure comment stays as-is.

pinin4fjords (approved):

PR #223, opened 2026-07-31
Add per-sample UMI handling for mixed experiments

### LINKED ISSUE #146: riboseq(with umi) rnaseq(without umi)
### Description of the bug

My riboseq data has umi, but rnaseq doesn't. How should I represent it? -- with_umi seems to be a global configuration

```console

```

### Relevant files

_No response_

### System information

_No response_

### LINKED ISSUE #149: Add test coverage for UMI dedup pathway
## Problem

PR #145 updated the `bam_dedup_umi` subworkflow from nf-core/modules, which added a third argument to the `SAMTOOLS_SORT` call (`index_format`). However, the `samtools/sort` module itself was not updated at the same time, so it still only declared 2 inputs. This caused a runtime error:

> Process `NFCORE_RIBOSEQ:RIBOSEQ:BAM_DEDUP_UMI:SAMTOOLS_SORT` declares 2 inputs but was called with 3 arguments

This was not caught by CI because no test profile exercises the `--with_umi true` code path.

Fixed in #148.

## Proposed fix

Add a test profile (or extend an existing one) that runs with `--with_umi true` so that the `BAM_DEDUP_UMI` subworkflow is exercised in CI. This would catch input cardinality mismatches and other regressions in the UMI dedup pathway.

pinin4fjords:
## Summary

- add an optional samplesheet `with_umi` field that overrides the global `--with_umi` fallback per sample
- partition UMI and non-UMI samples in pipeline orchestration and run the unmodified nf-core preprocessing subworkflow under separate aliases
- extract and deduplicate UMIs only for enabled samples across the primary and hybrid alignment paths, then recombine both sample groups downstream
- keep UMI routing out of nf-core component metadata and avoid component patches
- document global, mixed-sample, override, pre-extracted UMI, and discard-read usage patterns
- cover an explicit per-sample UMI override and a blank-field global fallback in one mixed end-to-end test

## Tests

- mixed UMI override and fallback nf-test with Docker and Nextflow 25.10.4
- mixed UMI integration run with Bowtie2 rRNA removal and automatic strandedness inference on Docker and Nextflow 26.04.6
- default trimming-enabled pipeline nf-test with Docker and Nextflow 25.10.4
- `prek run --all-files`

Closes #146
Closes #149


FelixKrueger (commented):
Read through this and traced the UMI routing on every path — the logic holds up, including the two-level guard (`any_sample_with_umi` deciding whether the dedup subworkflows exist, `.filter { meta.with_umi }` deciding which samples reach them), raw FastQC still running before the branch, and the empty-channel case where UMI samples only exist among the RNA-seq libraries. Docs are good, and the negative assertion in the mixed test (`dedup_outputs.every { !it.contains('SRX11780885') }`) checks the thing that actually matters. Three things I'd like before merge: document the new `meta.with_umi` input contract in both patched subworkflows' `meta.yml` (the `.diff`s currently record it as unchanged, so the contract is undiscoverable) plus a note on whether these patches get upstreamed; a CHANGELOG line warning that stamping `with_umi` into `meta` invalidates existing `-resume` caches for all users, UMI or not; and a test for `umi_discard_read` combined with the mixed samplesheet, which is the riskiest untested interaction here. Everything else below is optional cleanup.

<details>
<summary>Detail on the above, plus smaller findings (AI-assisted review — bin freely)</summary>

### Upstreaming the subworkflow patches

`fastq_fastqc_umitools_fastp` and `fastq_fastqc_umitools_trimgalore` are shared with rnaseq and others, and rnaseq wants this same feature. Every future `nf-core subworkflows update` will now conflict. The patch is already written in the right shape for upstreaming — `meta.with_umi == null ? with_umi : meta.with_umi` is backwards compatible when the key is absent — so it should port cleanly. Worth noting that null branch is unreachable in this pipeline, since `resolveSampleUmi()` eagerly stamps a Boolean onto every meta. That's the correct choice (the dedup filters need the key present), but it means the fallback is dead and untested locally; a component-level nf-test upstream would cover it.

### `resolveSampleUmi()` defensive branches

With `"type": "boolean"` in the samplesheet schema, nf-schema only ever yields `true`, `false` or `null`, so `instanceof List`, `== ''` and the `toString().toBoolean()` coercion are all unreachable. The coercion is the one I'd change: Groovy's `String.toBoolean()` returns true only for `"true"`, so if the schema is ever loosened, `with_umi,1` or `with_umi,yes` resolves to **`false`** — a wrong answer rather than an error, on a setting that materially changes read counts. Since the schema is the guarantee:

```groovy
def resolveSampleUmi(meta, default_with_umi) {
    meta + [ with_umi: (meta.with_umi == null ? default_with_umi : meta.with_umi) as boolean ]
}
```

Or keep the coercion but `error()` on unrecognised values instead of falling through.

### Fourth samplesheet parse

`samplesheetToList` now runs four times per launch: once in `PIPELINE_INITIALISATION`, then three times in `RIBOSEQ` (channel factory at `:146`, `samplesheetNeedsSalmonForStrandedness` at `:169`, `samplesheetHasUmi` at `:174`). Because `fastq_1`/`fastq_2` carry `"exists": true`, each parse re-stats every FASTQ — a network round-trip per file per parse for remote samplesheets. Three were pre-existing, but collapsing them is cheap while you're in here:

```groovy
def samplesheet_rows    = samplesheetToList(params.input, "${projectDir}/assets/schema_input.json")
def any_sample_with_umi = samplesheet_rows.any { meta, _f1, _f2 -> resolveSampleUmi(meta, params.with_umi).with_umi }
salmon_index_available  = (params.salmon_index as boolean) || samplesheet_rows.any { meta, _f1, _f2 -> meta.strandedness == 'auto' }
channel.fromList(samplesheet_rows).map { ... }
```

### `branch` instead of twelve `.filter`s

`workflows/riboseq/main.nf` repeats `.filter { meta, _x -> meta.with_umi }` / `!meta.with_umi` twelve times across three near-identical blocks. Nothing structurally keeps the two predicates complementary, and the `.mix()` back is load-bearing — omitting it silently drops non-UMI samples from the rest of the pipeline. `branch` states the partition once and avoids traversing `ch_genome_bam` twice:

```groovy
ch_genome_bam.branch { meta, _bam -> umi: meta.with_umi ; no_umi: true }.set { ch_genome_bam_by_umi }
```

Also worth an inline comment: the two hybrid blocks mix back only the transcriptome BAM and deliberately discard the genome dedup output, which next to the primary block mixing all three channels reads like an omission.

### `conf/modules.config:862` is now tautological

`BAM_DEDUP_STATS_SAMTOOLS_UMI*:SAMTOOLS_INDEX` only ever receives UMI-filtered BAMs, so `meta.with_umi` is invariably true and the `saveAs` collapses to "always publish". The change does fix a real gap — with `--with_umi false` and samplesheet opt-in, dedup ran but the index went unpublished — but it now reads like per-sample gating when it isn't. Either drop the term with a one-line comment or note why it stays. The sibling change at `:522` genuinely does vary per sample and is correct as written.

### Other test-coverage gaps

- Test 2's samplesheet has a single row, so `umi_outputs.every { it.startsWith('SRX11780885') }` is trivially satisfied — it shows extraction happened but not that the fallback applies *per sample*. Two rows with no `with_umi` column would test it.
- Nothing exercises the new `error()` at `:277` for a multi-run sample with conflicting `with_umi` values.
- The documented `--skip_umi_extract` + `with_umi=true` combination is untested. The logic does hold (the extract block is skipped wholesale, dedup keys independently off `meta.with_umi`), but the docs make a specific promise that nothing pins down.
- `tests/tags.yml` triggers on `**.nf`, `**.config`, `**.nf.test`, `**.json` but not `**.csv`, so a change to only `tests/samplesheet_mixed_umi.csv` won't run the pipeline tests. Pre-existing, but this is the first PR to add CSV fixtures the tests depend on.

Skipping snapshots in favour of filename assertions looks like the right call given this repo's output nondeterminism.

### Nits

- CHANGELOG entry sits under `Fixed`; adding a samplesheet column and a new capability reads more like `Added`.
- The two new `include` lines pad the braces (`{ resolveSampleUmi      }`) unlike their neighbours.
- No security concerns — `with_umi` is schema-constrained and only ever used as a channel predicate, never interpolated into a command.
- Performance is a net win outside the parse issue: non-UMI samples now skip `UMITOOLS_EXTRACT` and `umi_tools dedup` entirely, and `directional` dedup is one of the most expensive steps in the pipeline.

</details>

FelixKrueger (approved):
CI is green on `1c8524d` — 34 checks passing across both Nextflow versions, the only skip being `Run AWS full tests`, which doesn't fire on PRs. That plus the read-through above is enough for me to approve; the routing logic held up on every path I traced.

Not blocking, but so they don't get lost at merge, the three asks from my earlier comment are the ones CI can't speak to: documenting the new `meta.with_umi` input contract in both patched subworkflows' `meta.yml` (with a note on whether these get upstreamed), a CHANGELOG line warning that stamping `with_umi` into `meta` invalidates existing `-resume` caches for everyone rather than just UMI users, and a test for `umi_discard_read` against the mixed samplesheet — that last one being the interaction where UMI samples get rewritten to `single_end: true` and then mixed back into a channel alongside genuinely paired-end non-UMI samples.

pinin4fjords:
Thanks @FelixKrueger , I'll update subworkflows in nf-core/modules before I merge this

PR #225, opened 2026-08-05
Restore the hybrid UMI dedup config lost to subworkflow aliasing

FelixKrueger:
`BAM_DEDUP_UMI` is included three times (plain, `_HYBRID`, `_HYBRID_RNA`), but `SAMTOOLS_SORT` and `UMITOOLS_PREPAREFORRSEM` are configured by a selector naming only the plain alias, so the hybrid aliases inherit neither directive. `UMITOOLS_PREPAREFORRSEM` then falls back to prefix `${meta.id}` while its input BAM — from an equally unprefixed `SAMTOOLS_SORT` — is already `${meta.id}.bam`, and the module's own guard aborts the run with *"Input and output names are the same, use `task.ext.prefix` to disambiguate!"*. `SAMTOOLS_SORT` separately loses its `-n`, so the step the subworkflow calls "restore name sorting" coordinate-sorts instead; fixing only the prefix would trade the abort for a silent correctness problem. This gives each hybrid alias its own directives. Behaviour is unchanged for the plain alias and for any run that does not reach the hybrid RNA path.

Hit on a real genome-scale AWS Batch run with `--with_umi --run_price --run_rpbp --extended_orf_analysis` once `--contrasts` was supplied.

<details>
<summary>Verification notes (AI-assisted)</summary>

**Why no test caught it.** The failing process is reached only with `--with_umi` *and* a novel-transcript source *and* `--contrasts` — `BAM_DEDUP_UMI_HYBRID_RNA` exists solely to requantify RNA against the hybrid transcriptome for the ORF-level DTE denominator, so without contrasts the alias is never instantiated. `-profile test` never sets `--with_umi`, and no pipeline test combines all three.

**Why only the RNA alias aborts.** `bam_dedup_umi/main.nf:107` calls `UMITOOLS_PREPAREFORRSEM` on the `paired_end` branch alone. Ribo-seq is single-end, so `BAM_DEDUP_UMI_HYBRID`'s branch is empty and its identical missing prefix has never surfaced. Its `SAMTOOLS_SORT` does run, though, which is why the sort fix is applied to both aliases while the prefix fix is applied only where the process actually runs — no dead config.

**Selector resolution, verified rather than assumed.** Minimal pipeline on Nextflow 25.10.4, subworkflow aliased three times as here, selectors in the same shape and order as the patch:

```
RIBOSEQ:BAM_DEDUP_UMI_HYBRID_RNA:SAMTOOLS_SORT           | args=-n | prefix=RNA.namesorted
RIBOSEQ:BAM_DEDUP_UMI_HYBRID_RNA:UMITOOLS_PREPAREFORRSEM |         | prefix=RNA.filtered
RIBOSEQ:BAM_DEDUP_UMI_HYBRID:SAMTOOLS_SORT               | args=-n | prefix=HYB.namesorted
RIBOSEQ:BAM_DEDUP_UMI_HYBRID:UMITOOLS_PREPAREFORRSEM     |         | prefix=<NO-PREFIX>
RIBOSEQ:BAM_DEDUP_UMI:SAMTOOLS_SORT                      | args=-n | prefix=PLAIN.namesorted
RIBOSEQ:BAM_DEDUP_UMI:UMITOOLS_PREPAREFORRSEM            |         | prefix=PLAIN.filtered
```

Input and output prefixes now differ, so the guard no longer fires. The new selectors are declared after each `.*:BAM_DEDUP_UMI_HYBRID*:.*` catch-all so they win on declaration order within the same specificity class, and the catch-alls' `publishDir` still applies because the override is per-directive rather than whole-block — both confirmed in the same run.

**Same root cause as #220.** That fixed `BAM_DEDUP_UMI`'s MultiQC channel; this fixes its `ext.prefix` and `ext.args`. Both come from configuring one alias of a thrice-aliased subworkflow by name. A catch-all such as `.*:BAM_DEDUP_UMI(_HYBRID(_RNA)?)?:SAMTOOLS_SORT` would prevent a fourth alias repeating it, but that is a larger change than this fix and would alter the plain alias's resolution, so it is left alone here.

</details>


conf/modules.config
```diff
@@ -1476,6 +1487,11 @@ process {
         ]
     }
 
+    withName: '.*:BAM_DEDUP_UMI_HYBRID:SAMTOOLS_SORT' {
```
pinin4fjords, line 1494: [medium] Paired-end Ribo-seq is unusual, but the pipeline currently accepts and documents it. That path reaches `UMITOOLS_PREPAREFORRSEM`; without an alias-specific prefix it and the inner sort can both publish `${meta.id}.bam` when intermediates are saved, so completion order decides which survives.

```suggestion
    withName: '.*:BAM_DEDUP_UMI_HYBRID:UMITOOLS_PREPAREFORRSEM' {
        ext.prefix = { "${meta.id}.umi_dedup.hybrid_transcriptome.filtered" }
    }

    withName: '.*:BAM_DEDUP_UMI_HYBRID:SAMTOOLS_SORT' {
```

conf/modules.config
```diff
@@ -746,6 +746,17 @@ process {
         publishDir = [ enabled: false ]
     }
 
+    // prepare-for-rsem needs name-sorted input and a prefix distinct from its own
+    // input; both are set per alias, so each alias needs its own.
+    withName: '.*:BAM_DEDUP_UMI_HYBRID_RNA:SAMTOOLS_SORT' {
```
pinin4fjords, line 751: [medium] Please exercise the failing path in `tests/stringtie_extended.nf.test`. The default test profile already supplies paired-end RNA and contrasts, and this test already enables the required novel/extended path:

```groovy
with_umi = true
umitools_bc_pattern = 'NNNN'
```

This keeps the RNA paired and reaches `BAM_DEDUP_UMI_HYBRID_RNA`. GitHub cannot make this an Apply suggestion until the test file is added to the PR diff.

pinin4fjords (changes requested):
AI-assisted review (Codex, on behalf of @pinin4fjords).

conf/modules.config
FelixKrueger: You're right, and my reasoning was wrong: I left it unprefixed because Ribo-seq is single-end, but `assets/schema_input.json` requires only `sample`, `fastq_1`, `strandedness` and `type` and permits `fastq_2`, so nothing rejects a paired-end Ribo-seq row. (Small correction to the framing: the docs don't document it as supported — `usage.md:580` states Ribo-seq libraries *are* single-end — which if anything makes it worse, since the path is reachable while nothing warns you.)

Added `.*:BAM_DEDUP_UMI_HYBRID:UMITOOLS_PREPAREFORRSEM`. Your collision point also sent me to the inner sort: `.*:BAM_DEDUP_UMI_HYBRID:SAMTOOLS_SORT` matches only the direct invocation, so `BAM_SORT_STATS_SAMTOOLS:SAMTOOLS_SORT` was unprefixed on both hybrid aliases and was the other half of the `${meta.id}.bam` race. Prefixed on both.

Verified on Nextflow 25.10.4 with the subworkflow aliased twice and the inner level present — all six instances resolve distinctly, and the direct/inner selectors don't cross-match:

```
HYBRID:SAMTOOLS_SORT                            | args=-n   | prefix=HYB.namesorted
HYBRID:BAM_SORT_STATS_SAMTOOLS:SAMTOOLS_SORT    | args=none | prefix=HYB.innersorted
HYBRID:UMITOOLS_PREPAREFORRSEM                  |           | prefix=HYB.filtered
HYBRID_RNA:SAMTOOLS_SORT                        | args=-n   | prefix=RNA.namesorted
HYBRID_RNA:BAM_SORT_STATS_SAMTOOLS:SAMTOOLS_SORT| args=none | prefix=RNA.innersorted
HYBRID_RNA:UMITOOLS_PREPAREFORRSEM              |           | prefix=RNA.filtered
```

I went with the prefix rather than rejecting paired-end Ribo-seq, since rejecting it is a user-facing behaviour change that seems worth its own PR.

conf/modules.config
FelixKrueger: Added `tests/umi_extended_orf.nf.test` — `--with_umi` plus `--skip_stringtie false` plus `--extended_orf_analysis`, with `-profile test` supplying the contrasts. That combination is the only one that instantiates `BAM_DEDUP_UMI_HYBRID_RNA`.

It asserts rather than snapshots: without the prefixes the run aborts, so `workflow.success` *is* the regression, and it additionally fails if an unprefixed `${meta.id}.bam` appears in `alignment/hybrid_star/deduplicated`. That keeps it useful without adding a 400-line snapshot whose recording would need a separate CI cycle.

On the skipped matrix — that turned out to be a gap worth fixing rather than a one-off: `nf-test.config`'s `triggers` lists `conf/test.config` but not `conf/modules.config`, so a change confined to it selects no tests at all. The matrix job never instantiated (`total_shards=0`) and `confirm-pass` went green over an empty matrix. I've added `conf/modules.config` to `triggers` — without it the new test wouldn't run for a config-only change either. It does mean config-only PRs now cost a full matrix, which seems right given that file controls every `publishDir` and `ext.args` in the pipeline.

conf/modules.config
pinin4fjords: Thanks, these selectors cover the missing aliases. One correction: the docs are contradictory. The [samplesheet example explicitly supplies fastq_2 for type=riboseq](https://github.com/nf-core/riboseq/blob/ef594901389d2226b02c01e71d96431e9ed75c87/docs/usage.md#L24-L27), while [the kallisto section says Ribo-seq libraries are single-end](https://github.com/nf-core/riboseq/blob/ef594901389d2226b02c01e71d96431e9ed75c87/docs/usage.md#L578-L581). So I don’t think we can treat paired-end input as unsupported unless validation explicitly rejects it. Resolved.

conf/modules.config
pinin4fjords: Medium: Thanks, the coverage itself looks good. The test currently fails in RiboWaltz. 

This test [applies an artificial NNNN UMI pattern to non-UMI data](https://github.com/nf-core/riboseq/blob/ef594901389d2226b02c01e71d96431e9ed75c87/tests/umi_extended_orf.nf.test#L12-L24), which leaves one sample without enough signal for RiboWaltz to estimate offsets. Could you add skip_ribowaltz = true here? The underlying cryptic failure on sparse data is worth addressing separately, but I don’t think it should expand the scope of this PR.

conf/modules.config
FelixKrueger: You're right and my correction was wrong — `docs/usage.md:25-27` shows `type=riboseq` rows with `fastq_2` populated, so the docs do present paired-end Ribo-seq as valid input. I'd only read the kallisto caveat. Agreed that it can't be treated as unsupported while validation accepts it, which makes the prefix the right fix rather than a rejection.

tests/umi_extended_orf.nf.test
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        cleanup {
+            if (System.getenv("NFT_PROFILE")?.contains("docker")) {
+                "docker system prune -a -f --volumes".execute()
+            }
+            new File("${launchDir}").deleteDir()
+        }
+    }
+}
```
pinin4fjords, line 108: Sorry, I should have spotted this in the first pass.

**Medium:** This test needs to follow the conventional pipeline-test structure used elsewhere, for example in [`stringtie_extended.nf.test`](https://github.com/nf-core/riboseq/blob/ef594901389d2226b02c01e71d96431e9ed75c87/tests/stringtie_extended.nf.test#L1-L55). The `stable_name` snapshot captures missing or colliding prefixes, so the bespoke filesystem assertions duplicate what the snapshot already records. Please also generate the corresponding snapshot file after applying this.

```suggestion
nextflow_pipeline {

    name "Test pipeline with --with_umi, a novel-transcript source and ORF-level DTE"
    script "../main.nf"
    tag "pipeline"
    tag "pipeline_riboseq"
    tag "extended_orf_analysis"
    tag "umi"

    test("-profile test --with_umi true --skip_stringtie false --extended_orf_analysis true --skip_ribowaltz true") {

        when {
            params {
                outdir                = "$outputDir"
                with_umi              = true
                umitools_bc_pattern   = 'NNNN'
                save_umi_intermeds    = true
                skip_stringtie        = false
                extended_orf_analysis = true
                skip_ribowaltz        = true
            }
        }

        then {
            def stable_name = getAllFilesFromDir(params.outdir, relative: true, includeDir: true, ignore: ['pipeline_info/*.{html,json,txt}'])
            def stable_path = getAllFilesFromDir(params.outdir, ignore: ['alignment/hybrid_star/deduplicated/**'], ignoreFile: 'tests/.nftignore_orf')

            assertAll(
                { assert workflow.success },
                { assert snapshot(
                    removeNextflowVersion("$outputDir/pipeline_info/nf_core_riboseq_software_mqc_versions.yml"),
                    stable_name,
                    stable_path
                ).match() }
            )
        }

        cleanup {
            if (System.getenv("NFT_PROFILE")?.contains("docker")) {
                "docker system prune -a -f --volumes".execute()
            }
            new File("${launchDir}").deleteDir()
        }
    }
}
```

<sub>AI-assisted review.</sub>

FelixKrueger:
The restructured test can't pass on the current test data, and I don't think params can fix it — could you take a look at the setup? Recording the snapshot failed twice: first in gene-level `ANOTA2SEQ_ANOTA2SEQRUN`, then, after switching the test to `translational_efficiency_method = 'dotseq'` to skip the gene-level fit, in `DOTSEQ_DOTSEQ_ORF` with `No data after filtering. Please check the count table.` Both look like consequences of `umitools_bc_pattern = 'NNNN'` on data that carries no real UMIs — four bases is only 256 possible UMIs, so reads at the same locus collide and get collapsed as duplicates. Same underlying reason RiboWaltz needed `skip_ribowaltz = true`. The alias fix itself looks healthy: the second run published 30 distinctly-prefixed BAMs across the six samples, including `*.umi_dedup.hybrid_transcriptome.namesorted.bam`, so `ext.args = '-n'` is reaching the sort on the hybrid alias and the hybrid quantification completed. Would a longer UMI pattern be viable here, or is there test data with real UMIs? Lengthening it should reduce the collisions but trims more off already-short reads, so I didn't want to guess at it.

<details>
<summary>Evidence and reasoning (AI-assisted)</summary>

**The bind.** `BAM_DEDUP_UMI_HYBRID_RNA` is only instantiated with `--with_umi` *and* a novel source *and* `--contrasts` (`workflows/riboseq/main.nf:768-799`). `--contrasts` in turn forces at least one TE method, since `translational_efficiency_method` must match `^(anota2seq|deltate|dotseq)(,...)*$`. So every configuration that reaches the alias also runs a TE method over the deduplicated counts, and so far each method fails on them.

**Failure 1** — run `31128453105`, 62 min, gene-level:

```
1359 genes were removed due to having 0 counts in at least one sample.
Error in anota2seqNormalize(tmpdataP = tmpNoZero$dataP, tmpdataT = tmpNoZero$dataT,  :
  object 'returndataP' not found
Calls: do.call ... anota2seqRNAseqPreProcessing -> anota2seqNormalize
```

**Failure 2** — run `31173758047`, 74 min, after `translational_efficiency_method = 'dotseq'`:

```
Error in DOTSeq(datasets = d, modules = modules, target = opt$target_level,  :
  No data after filtering. Please check the count table.
```

`dotseq` was chosen because it is ORF-level only, so it skips the gene-level fit while leaving the alias path intact — the gating conditions above carry no `te_methods` term. That worked as intended; the failure simply moved downstream. `tests/dotseq.nf.test` passes on the same data at full depth, so DOTSeq is fine here and `--with_umi` is the only differing variable.

**What the fix demonstrably does.** Five distinctly-prefixed outputs per sample, 30 across six samples, no `${meta.id}.bam` collision:

```
SRX11780885.umi_dedup.hybrid_transcriptome.namesorted.bam
SRX11780885.umi_dedup.genome.sorted.bam
SRX11780885.umi_dedup.transcriptome.sorted.bam
SRX11780885.hybrid_transcriptome.sorted.bam
SRX11780885.primary.bam
```

`namesorted` is the one worth noting — that is the `-n` half of the defect, which a crash would never have surfaced. `quantification/salmon_hybrid`, `quantification/orf_level` and `quantification/inframe_psite` all completed.

**No snapshot has been committed.** nf-test does write a `.snap` on these runs, because snapshot creation and `workflow.success` are separate assertions in the same `assertAll` — but it records a truncated output tree (no DTE outputs, pending tasks cancelled), so committing it would pin a baseline that asserts less than it should and then pass quietly.

</details>

pinin4fjords:
Thanks. I don't think lengthening the fake UMI is reliable: those bases come from the biological insert, so reads at the same locus tend to receive the same pseudo-UMI however long it is. We're trying a focused alias-level test on the VM using the existing real-UMI BAM fixtures instead. That should exercise the selectors and prefixes without dragging the full TE stack into a config regression test. Will report back.

FelixKrueger:
CI is green on `9a169c1` — 34/34, `latest-everything` included. Thanks for sorting out the MultiQC instability, and for scoping the ignore to just this test rather than globally.

Your `CHANGES_REQUESTED` from 08-06 is still the operative review, and a green run won't clear it on its own, so this needs a re-review from you when you get a chance.

One thing to disregard while you're looking: the ❌ nf-test sticky comment above is stale. It's pinned to `docker | latest-everything | Shard 13/14`, which passes now — the poster only writes that comment on failure and never clears it afterwards (nf-core/tools#4436).

pinin4fjords (approved):
Looks good now. The requested fixes are covered and CI is green.

AI-assisted review (Codex, on behalf of @pinin4fjords).

PR #226, opened 2026-08-10
Make orf_class purely positional and add uoORF, doORF and intORF

FelixKrueger:
`orf_class` currently mixes position with length: a short ORF is labelled `smORF` wherever it sits relative to the CDS, and the CDS-overlap and internal distinctions the callers already report get folded into `uORF`/`dORF`/`other`. This makes `orf_class` purely positional, adds `uoORF`, `doORF` and `intORF`, and moves length onto a separate `is_smorf` column controlled by `--smorf_max_aa` (default 100, which reproduces the previous `smORF` membership exactly). It also adds `orf_type_native`, carrying each caller's own label so every harmonisation decision stays auditable. Breaking for anything reading `orf_class`; the CHANGELOG gives both migration directions.

The module side is already merged — nf-core/modules#12498, #12599 and #12613 — and `modules.json` pins them.

<details>
<summary>Design notes and verification (AI-assisted)</summary>

**Why class had to leave orfmerge's grouping key.** Callers disagree about class for the same ORF, and Ribo-TISH structurally cannot agree: `tisType()` tests only the start position, so its `5'UTR` covers uORF ∪ uoORF. Keying the merge on class would emit one row per disagreeing caller. The rule this settled on: a class is safe to add only if either every caller can positively report it, or the merge is class-agnostic.

**Two clustering fixes came along with it.** Reciprocal overlap is now measured on summed exon-block intersection rather than the outer genomic span, which for a spliced ORF is mostly intron; and clustering uses complete linkage, so a chain of partial overlaps cannot fold distinct ORFs into a single row.

**Row counts move in both directions**, which the CHANGELOG spells out. Down: ORFs whose callers disagreed on class now merge. Up: a short truncated CDS variant is no longer folded into its transcript's full-length CDS, and short transcript-anchored ORFs are keyed on their exact span.

**Deliberately not included.** No `altCDS` class (RiboCode folds variants into `annotated` and ribotricer into its `internal` fall-through, so it would fail the safety rule); ribotricer's `internal` stays `other` because it is that tool's terminal fall-through rather than a frame-tested call; and stop-codon keying for short transcript-anchored ORFs is documented and pinned by a test rather than fixed.

**Snapshots.** The four ORF-catalogue snapshots were re-recorded on the versions this repo's CI pins (nf-test 0.9.4, Nextflow 25.10.4). Every md5 that moved belongs to a file carrying `orf_class`; the `ribotish.normalised.bed12` md5s did not move, which is what a classification-only change should look like.

</details>


CHANGELOG.md
```diff
@@ -81,6 +84,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- [#226](https://github.com/nf-core/riboseq/pull/226) - **Breaking:** `orf_class` values change for ORFs the callers place relative to the CDS. RiboCode's and ribotricer's `Overlap_uORF`/`Overlap_dORF` were being folded into `uORF`/`dORF` by substring matching and now map to `uoORF`/`doORF`; RiboCode's and Ribo-TISH's `internal` and PRICE's `iORF` now map to `intORF` rather than `other`; PRICE's `uoORF` maps to `uoORF` rather than `uORF` and its `orphan` to `novel_u` rather than `other`. ribotricer's `internal` stays `other` because it is that tool's fall-through rather than a frame-tested call. To keep a pre-existing filter's meaning, read the old `uORF` as the new `uORF` ∪ `uoORF`, the old `dORF` as `dORF` ∪ `doORF`, and note that the old `other` no longer holds the internal ORFs. Anything consuming `orf_class` — including the ORF-level DOTSeq analysis, which matches `uORF`/`dORF` literally — sees the new vocabulary ([@FelixKrueger](https://github.com/FelixKrueger))
```
suhrig, line 88: > Anything consuming `orf_class` — including the ORF-level DOTSeq analysis, which matches `uORF`/`dORF` literally — sees the new vocabulary

Just so we are clear: This means that DOTSeq will now consider a reduced set of ORFs on [this line](https://github.com/nf-core/riboseq/blob/42fa9c222f31f2f44aaa76e51b8a4b08298a6217/modules/nf-core/dotseq/dotseq/templates/dotseq.R#L422) compared to the previous approach. I'm not saying this is worse. I just want to rule out an oversight.

suhrig (approved):
Great enhancement! In particular, the complete linkage mapping is biologically so much more meaningful than the previous approach!

CHANGELOG.md
FelixKrueger: Worth checking, thanks. That heatmap call is the only place orf_class reaches DOTSeq -- sorf_type appears nowhere else in the template -- so the DTE and DOU statistics and the set of ORFs they test are unchanged; what shrinks is only that figure's input, since ORFs the old substring matching folded into uORF are now uoORF and no longer satisfy sorf_type = "uORF". Happy to add uoORF and doORF as further fallbacks in a follow-up if you would rather the heatmap kept its previous coverage.

suhrig: I don't mind. One could even argue that it is better now, since those are the true *upstream* ORFs. Zero overlap.

pinin4fjords, line 88: Can we try to keep changelog entries a bit less verbose? Claude does tend to get carried away, they should only a be a sentence.

pinin4fjords (approved):

CHANGELOG.md
FelixKrueger: Fair -- two of those had grown into full paragraphs with migration instructions inline. #227 trims all seven to a single sentence each, 440 words down to 136; the vocabulary and migration detail is already in docs/output.md and docs/usage.md, which is where it belongs.

PR #228, opened 2026-08-11
Let nf-test report on docs-only PRs

FelixKrueger:
`nf-test.yml` skips itself on docs-only changes via `paths-ignore`, but `confirm-pass` lives inside that workflow and is a required check — so a docs-only PR can never report it and sits `BLOCKED` even when approved and fully green. #227 is stuck that way now: it changes only `CHANGELOG.md`, which matches `**/*.md`. Removing the filter lets the workflow start and select nothing, which costs two trivial jobs and lets `confirm-pass` report.

No test coverage changes: test selection is done by `nf-test-changes` using `--changed-since`, not by the trigger.

<details>
<summary>Why this reports success rather than running the suite (AI-assisted)</summary>

The short-circuit already exists in the workflow:

- `nf-test-changes` computes `total_shards`; for a docs-only diff it is `0`.
- `nf-test` carries `if: ${{ needs.nf-test-changes.outputs.total_shards != '0' }}`, so it is **skipped**.
- `confirm-pass` is `if: always()` and fails only on `contains(needs.*.result, 'failure')` or `'cancelled'`. `skipped` is neither, so it passes.

This was already observed on nf-core/riboseq#225, whose first run selected zero tests and reported `confirm-pass` success over an empty matrix.

Worth noting the same `paths-ignore` block is in the nf-core pipeline template, so any pipeline with `confirm-pass` as a required check has this deadlock for docs-only PRs.

</details>

<details>
<summary>Why <code>types:</code> appears (AI-assisted)</summary>

The `actions_nf_test` lint rule documents the expected trigger as a bare `pull_request:`, then fails when that subtree parses to `None` — which is what a bare key gives. Naming GitHub's own default activity types (`opened`, `synchronize`, `reopened`) satisfies the check without changing when the workflow runs. The rule contradicting its own documentation looks like an nf-core/tools bug worth fixing separately.

</details>


.github/workflows/nf-test.yml
```diff
@@ -1,12 +1,7 @@
 name: Run nf-test
 on:
   pull_request:
-    paths-ignore:
-      - "docs/**"
-      - "**/meta.yml"
-      - "**/*.md"
-      - "**/*.png"
-      - "**/*.svg"
+    types: [opened, synchronize, reopened] # GitHub's defaults; named so the nf-core lint sees a non-empty subtree
```
pinin4fjords: Let's keep GitHub's default event set implicit and disable `actions_nf_test` for this intentional deviation, as rnaseq does.

```suggestion
```

pinin4fjords (changes requested):
I think we should port the complete fix from nf-core/rnaseq#1618. As written, `confirm-pass` only needs `nf-test`, so if `nf-test-changes` fails, `nf-test` is skipped and the required check can pass without seeing the upstream failure.

Please also set `needs: [nf-test-changes, nf-test]` on `confirm-pass` and add `lint: actions_nf_test: false` to `.nf-core.yml`. That matches the established rnaseq solution without teaching the trigger to satisfy the lint rule.

.github/workflows/nf-test.yml
FelixKrueger: Ported in 44aa011 -- confirm-pass now needs both jobs, the lint deviation is declared in .nf-core.yml, the bare pull_request: is back, and there is a changelog entry. You are right about the hole: a failed nf-test-changes left nf-test skipped and the required check green over a crashed selector, which is the same shape as the bug this PR fixes. One aside, take it or leave it: the actions_nf_test rule rejects the bare form its own docstring documents, so I opened nf-core/tools#4441 -- if that lands the exception here becomes unnecessary, but the exception is the right call either way.

pinin4fjords (approved):

PR #229, opened 2026-08-12
Give CUSTOM_ORFMERGE enough memory and wall-clock at genome scale

FelixKrueger:
`CUSTOM_ORFMERGE` does not fit `process_medium` on a genome-scale cohort. On an 18-sample, 4-caller run it was killed at 36 GB/8h and again at 72 GB/16h before completing in 14h42m at 108 GB — about 39 hours of wall-clock to do 15 hours of work. This sets `memory` and `time` on the selector that already governs the process, and leaves `cpus` on the label because `orfmerge.py` is single-threaded. Resource directives are not part of a task's cache key, so this is safe to apply to a running session with `-resume`.

<details>
<summary>Detail (AI-assisted)</summary>

- Both kills report exit 143 with an empty `.command.err`, which reads as OOM. The durations matching `8.h * task.attempt` to the second are what identify the wall-clock limit as the terminator instead.
- `time` is generous (30h, matching the `RPBP_ESTIMATE*BAYESFACTORS` override in the same file) because it is a ceiling billed on actual runtime, whereas `memory` is a reservation and stays at the measured 108 GB.
- `process_high` would be the wrong reflex here: its first attempt is 72 GB/16h, exactly the configuration that already failed.
- Attempt 3 succeeded in less time than attempt 2 was allowed, so the extra memory bought speed rather than only headroom.
- Cache-neutrality checked on a live resume rather than argued: 3123 of 3123 banked tasks still cached with the change applied.

</details>

### DIFF
--- a/conf/modules.config
+++ b/conf/modules.config
@@ -1108,9 +1108,8 @@ process {
     }
     withName: '.*:ORFTABLE_FASTA_GTF_BUILDORFCATALOGUE:CUSTOM_ORFMERGE' {
         ext.args   = { "--min-callers ${params.orf_min_callers} --min-samples ${params.orf_min_samples}" }
-        // Complete-linkage clustering, ~15h and over 72 GB at genome scale; first-attempt headroom.
-        memory     = { 108.GB * task.attempt }
-        time       = { 30.h   * task.attempt }
+        // Cohort merge takes ~15h at genome scale; first-attempt headroom.
+        time       = { 30.h * task.attempt }
         // The published consensus is the merger's only when the collapse is
         // skipped; otherwise CUSTOM_ORFCOLLAPSE publishes the post-collapse one.
         publishDir = params.skip_orf_collapse ? [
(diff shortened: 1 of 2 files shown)


conf/modules.config
```diff
@@ -1108,6 +1108,9 @@ process {
     }
     withName: '.*:ORFTABLE_FASTA_GTF_BUILDORFCATALOGUE:CUSTOM_ORFMERGE' {
         ext.args   = { "--min-callers ${params.orf_min_callers} --min-samples ${params.orf_min_samples}" }
+        // Complete-linkage clustering, ~15h and over 72 GB at genome scale; first-attempt headroom.
+        memory     = { 108.GB * task.attempt }
```
mashehu: why 108? would go for something rounder and also closer to the tested outcome: 
```suggestion
        memory     = { 80.GB * task.attempt }
```

mashehu (approved):

conf/modules.config
FelixKrueger: good pushback -  I just took a look at the numbers again. It turns out that it only required 3 attempts because it ran out of time (and the numbers were 36->72->108). Another test run is currently under way, but likely we only need to increase the runtime limit and not memory at all (in fact memory was ~6GB only). CHeers

PR #230, opened 2026-08-14
Pin DESEQ2_DELTATE_ORF's cpus at genome scale

FelixKrueger:
`DESEQ2_DELTATE_ORF` inherits `process_medium`, whose `cpus` scale with `task.attempt`. The module hands `task.cpus` to `MulticoreParam`, so every retry forks more DESeq2 workers, roughly tripling peak memory without finishing any sooner — on a genome-scale cohort the same seven contrasts ran in 3h 16m at 6 cpus and 13–14h at 18. This pins `cpus = 6`, the configuration that completes. `time` keeps its existing attempt scaling and `memory` is unchanged, so nothing else about the process changes.

<details>
<summary>Measurements behind the change (AI-assisted analysis)</summary>

Two genome-scale runs of the same seven contrasts over identical input (`ORF_COUNT_MATRIX` wchar 15.8 MB in both):

| cpus | realtime | peak_rss | core-hours |
|---|---|---|---|
| 6 | 3h 15–16m | 17.9–19 GB | 42.5 |
| 18 | 13h 04m – 14h 22m | 55.1–57.8 GB | 312.2 |

`peak_rss` per cpu is 3.08 GB at 6 and 3.14 GB at 18, so memory demand tracks the worker count rather than the problem size. Core-hours grew 7.3× for the same work, which isolates the parallelism overhead from any scheduling effect.

Every failure in both runs was `exit 143` at the wall-clock ceiling, never `137`, so memory was never the binding constraint. The label's 36 GB covers the measured 17.9–19 GB and gets no override.

The relevant lines are in `modules/local/deseq2/deltate/templates/deseq2_deltate.R`: `cores = as.integer('$task.cpus')`, `if (opt$cores > 1) register(MulticoreParam(opt$cores))`, and three `DESeq(..., parallel = (opt$cores > 1))` calls.

</details>

### DIFF
--- a/conf/modules.config
+++ b/conf/modules.config
@@ -1257,6 +1257,8 @@ process {
             params.rna_lfc_threshold ? "--lfc_threshold_rna ${params.rna_lfc_threshold}" : '',
             params.ribo_lfc_threshold ? "--lfc_threshold_ribo ${params.ribo_lfc_threshold}" : ''
         ].join(' ').trim() }
+        // Fitting hundreds of thousands of ORFs exceeds 8h at genome scale; first-attempt headroom.
+        time       = { 16.h * task.attempt }
         publishDir = [
             path: { "${params.outdir}/translational_efficiency/orf_level/deltate" },
             mode: params.publish_dir_mode,
--- a/conf/modules.config
+++ b/conf/modules.config
@@ -1257,8 +1257,8 @@ process {
             params.rna_lfc_threshold ? "--lfc_threshold_rna ${params.rna_lfc_threshold}" : '',
             params.ribo_lfc_threshold ? "--lfc_threshold_ribo ${params.ribo_lfc_threshold}" : ''
         ].join(' ').trim() }
-        // Fitting hundreds of thousands of ORFs exceeds 8h at genome scale; first-attempt headroom.
-        time       = { 16.h * task.attempt }
+        // Fitting ~286k ORFs takes 6-14h at genome scale and varies with load; first-attempt headroom.
+        time       = { 24.h * task.attempt }
         publishDir = [
             path: { "${params.outdir}/translational_efficiency/orf_level/deltate" },
             mode: params.publish_dir_mode,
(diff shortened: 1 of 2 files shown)


FelixKrueger:
Drafting this: 16h is insufficient too — the retry was killed at 16h 0m 13s on 12 cpus. Attempt 3 is running with 24h, so I'll set the number from that measurement rather than guess again.

Worth flagging that the same seven contrasts completed in roughly 3h of compute on an earlier run over the same data, so I'd like to rule out contention between the concurrently running tasks before assuming more wall-clock is the right fix at all.

FelixKrueger:
Updated to 24h. On the 24h/18-cpu attempt three of the seven contrasts completed in 6h 05m, 13h 04m and 13h 04m, having been killed at 8h and again at 16h; peak RSS was 19 GB against the 36 GB the label already grants, so memory is not implicated.

Keeping it in draft until the remaining four finish tonight, since those are on the last retry and would falsify 24h if any of them hit it.

One caveat worth recording: the requirement is load-sensitive rather than fixed. The same seven contrasts finished in 3h 16m on an earlier run over input differing by 38 rows out of 285,952, and the fastest of the three above started later than the two slowest. That points at contention between the concurrent tasks — each scales its DESeq2 workers off `task.cpus`, so every retry escalation gives all seven more workers competing for the same hardware. If that is the mechanism then `maxForks` is the better lever and this ceiling is only buying completion; happy to look at that separately rather than widen this PR.

conf/modules.config
```diff
@@ -1257,6 +1257,9 @@ process {
             params.rna_lfc_threshold ? "--lfc_threshold_rna ${params.rna_lfc_threshold}" : '',
             params.ribo_lfc_threshold ? "--lfc_threshold_ribo ${params.ribo_lfc_threshold}" : ''
         ].join(' ').trim() }
+        // DESeq2 forks one worker per cpu, so peak memory tracks the cpu count.
```
mashehu: Is this comment really correct/necessary? This doesnt address Memory (PR description might cover it but falls under ai;dr)

mashehu (approved):

conf/modules.config
FelixKrueger: Removed in 061ba8c. You are right that it documented memory in a block that sets no memory directive, which is the confusion you spotted — the measurement lives in the commit message and PR body instead. Happy to trim the description too if it still reads long.

FelixKrueger: noted. Cheers!

PR #234, opened 2026-09-06
Dev -> master for v2.0.0

pinin4fjords:
## Release 2.0.0

**551 commits, 731 files changed** since v1.2.0.

This release adds the extended ORF analysis pathway end to end: a hybrid canonical + novel-transcript reference GTF (`--canonical_gtf`/`--novel_gtf`/StringTie), two new opt-in ORF callers (PRICE, Rp-Bp), a cross-caller cohort ORF catalogue with smORF collapse and consensus scoring, per-ORF P-site quantification, and ORF-level differential translation analysis. It also switches the default `--te_quantification_method` to `plastid_psite`, reworks rRNA removal and pseudo-alignment tool selection, reorganises pipeline outputs into category-based directories, and updates nf-core template/module dependencies throughout.

**Breaking changes** (see `Changed`/`Removed` below for full detail):

- `--te_quantification_method` default changed from `alignment` to `plastid_psite` (#160)
- `--skip_ribotricer` replaced by `--run_ribotricer` with inverted default (#163)
- Output directory layout reorganised into categories (#196)
- `orf_class`/smORF vocabulary changed; short ORFs now selected via `is_smorf` (#226)
- `--min_mapped_reads`, `--skip_pseudo_alignment`, `--skip_alignment` removed (#197)

### `Credits`

Special thanks to the following for their contributions to this release:

- [@bediagan](https://github.com/bediagan)
- [Evangelos Karatzas](https://github.com/vagkaratzas)
- [Felix Krueger](https://github.com/FelixKrueger)
- [Friederike Hanssen](https://github.com/FriederikeHanssen)
- [Ira A Iosub](https://github.com/iraiosub)
- [Jack Tierney](https://github.com/JackCurragh)
- [Jonathan Manning](https://github.com/pinin4fjords)
- [Luisa Santus](https://github.com/luisas)
- [Matthias Hörtenhuber](https://github.com/mashehu)
- [Maxime U Garcia](https://github.com/maxulysse)
- [Sebastian Uhrig](https://github.com/suhrig)

### `Added`

- [#125](https://github.com/nf-core/riboseq/pull/125) - Add rRNA removal tool selection with support for SortMeRNA (default), Bowtie2, and RiboDetector ([@pinin4fjords](https://github.com/pinin4fjords))
[...]

PR #235, opened 2026-09-06
fix: post-release-review cleanup

pinin4fjords:
## Summary

Fixes for bugs found while adversarially reviewing #234 (the v2.0.0 release PR) before merging it. Each fix is verified with module-level nf-test and a full pipeline-level nf-test run on real (chr20) data.

Two fixes are also merged upstream in nf-core/modules ([#12895](https://github.com/nf-core/modules/pull/12895), [#12896](https://github.com/nf-core/modules/pull/12896)); `modules.json` here points at those merged commits.

## What changed

- `--skip_gtf_transcript_filter`'s `ext.args` used `?:` instead of `? :`, so the check was skipped by default and crashed if explicitly enabled. Fixed.
- Gene-level `DESEQ2_DELTATE` passed `--samples_batch_col`, a flag its R script doesn't recognise (`--batch_col` does, and the ORF-level sibling already had it right). Any run with a batch column crashed.
- `deseq2_deltate.R` fit the interaction model on the whole sample sheet instead of subsetting to the two contrast levels, unlike anota2seq/DOTSeq. Harmless with only two levels, wrong with 3+ levels sharing a `contrast_variable`. Now subsets like the others do.
- `orf_count_matrix`: a sample with zero P-sites produced an empty file with no `sample_id`, so it vanished from the matrix instead of showing as zero-filled. Now takes the expected sample list explicitly (JSON+base64-encoded, so it survives commas/quotes in sample ids).
- `quantify_orf_psite`: simplified sample-id/TSV collection to one `.collect(flat: false)` pass instead of two separate channel subscriptions.
- `RIBOCODE_METAPLOTS` exits 1 on sparse periodicity data, aborting the whole run, even though `orf_caller_dispatch` already has warn-and-skip logic for exactly that case. Added `errorStrategy = 'ignore'` so it actually fires.
- Reverted the temporary GitHub-hosted CI runner swap back to the self-hosted fleet. GitHub's `ubuntu-latest` now blocks unprivileged user namespaces, which breaks Singularity outright.
- Two upstream nf-core/modules fixes (linked above), synced here.

## Test plan

- [x] Module-level nf-test for every fix
- [x] Full pipeline nf-test (`deltate`, `dotseq`, `stringtie_extended`, `umi_extended_orf`, `orf_catalogue_no_novel_source`, `novel_gtf`) on chr20 data
- [x] `default.nf.test` unaffected
- [x] Adversarial review pass on every fix

### DIFF
--- a/.github/workflows/nf-test.yml
+++ b/.github/workflows/nf-test.yml
@@ -21,7 +21,9 @@ env:
 jobs:
   nf-test-changes:
     name: nf-test-changes
-    runs-on: "ubuntu-latest" # temporarily using GitHub-hosted runners while self-hosted runners are down
+    runs-on: # use self-hosted runners
+      - runs-on=${{ github.run_id }}-nf-test-changes
+      - runner=4cpu-linux-x64
     outputs:
       shard: ${{ steps.set-shards.outputs.shard }}
       total_shards: ${{ steps.set-shards.outputs.total_shards }}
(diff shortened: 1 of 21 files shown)


JackCurragh (approved):
Looks good to me. Only think I questioned was whether the building of the orf count matrix should be validating the input but figured it'd be overkill to request changes for that.

pinin4fjords:
Thanks @JackCurragh !

PR #236, opened 2026-09-14
fix(quantify_inframe_psite_plastid): correct malformed singularity container URL

pinin4fjords:
## Summary

`QUANTIFY_INFRAME_PSITE_PLASTID`'s singularity container URL uses `==` instead of `:` as the tag separator (`bedtools==2.31.1--h13024bc_3`), which 404s against `depot.galaxyproject.org`. The docker branch of the same ternary is correct.

This has been broken since the module was introduced in January, but never caught: dev-branch CI never runs the `singularity` profile (only master-targeted PRs do), so this is effectively the first time this code path has been exercised under Singularity. It's currently breaking every `singularity` shard on #234 (the dev -> master release PR).

## What changed

- Fixed the URL: `bedtools==2.31.1--h13024bc_3` -> `bedtools:2.31.1--h13024bc_3`
- Widened `workflow.containerEngine == 'singularity'` to `in ['singularity', 'apptainer']`, matching the convention used everywhere else in the pipeline (same line, same class of bug waiting to happen under `-profile apptainer`)

## Test plan

- [x] Verified the corrected URL resolves (`curl -I` -> 200 OK, was 404 before)
- [x] `nf-core pipelines lint` / pre-commit hooks pass


pinin4fjords:
Admin-merged with required-checks bypass. 1 in-flight workflow cancelled: `Run nf-test`.

Rationale: `Run nf-test` never actually completed on this branch - each push (the container-URL fix, then two changelog-only follow-ups) cancelled the previous run's in-flight nf-test before it finished, so there was no green baseline to diff an "unverified" range against. User explicitly overrode waiting for a run: the fix is a one-character change to a previously-404ing singularity container URL (`bedtools==2.31.1...` -> `bedtools:2.31.1...`), independently verified by `curl -I` (404 before, 200 OK after) before pushing. The two follow-up commits touched only `CHANGELOG.md`.

PR #238, opened 2026-09-15
Drop dead versions.yml publishDir guards; relocate topic-versions collection

pinin4fjords:
## Summary
- Removes the `filename.equals('versions.yml') ? null : ...` `saveAs` guard from `conf/modules.config` selectors whose underlying module(s) now emit versions purely via `eval()` tuples on the `versions` topic channel and never write a `versions.yml` file, so the guard was dead code.
- Keeps the guard everywhere a module still writes a real `versions.yml` (template-based modules tagged onto the topic channel, and modules not yet migrated to topics), and on selectors whose regex covers a mix of migrated/unmigrated processes.
- Moves the `topic_versions`/`topic_versions_string` channel collection in `workflows/riboseq/main.nf` from the top of `main:` down to right before it's consumed (just ahead of the MultiQC block), per [review feedback on PR #234](https://github.com/nf-core/riboseq/pull/234#discussion_r4014047482).

## Test plan
- [x] `nextflow config .` parses `conf/modules.config` cleanly
- [x] `nextflow lint workflows/riboseq/main.nf` shows no new errors (the one pre-existing error is in an untouched template file, `subworkflows/nf-core/utils_nextflow_pipeline/main.nf`)
- [ ] CI


maxulysse (approved):
<3

